PR #89, opened 2020-09-22
Finishing finishing touches

abettermap:
Thought a good way to cap off the summer would be:

1. 3D fly-to's
1. And an in-panel Clear Filters btn

Enjoy!

### DIFF
--- a/src/components/legend/LayerLabelSelect.tsx
+++ b/src/components/legend/LayerLabelSelect.tsx
@@ -7,11 +7,7 @@ import { LangRecordSchema, LangSchemaCol } from '../../context/types'
 
 export const LayerLabelSelect: FC = () => {
   const { state, dispatch } = useContext(GlobalContext)
-  const labelFields: LangSchemaCol[] = [
-    'Language',
-    'Endonym',
-    'Language Family',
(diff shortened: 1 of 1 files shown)


rperlin-ela:
I am 100% loving the Clear Filters button, and the 3D is super-cool and hella shnazzy. I agree that it’s low-key and gentle. The Reset button may be a little subtle for people, with the compass looking a bit like the Mapbox logo below it, but I figured it out pretty instantly because of the way it flew in. Awesome additions!

abettermap:
great, glad ya dig.

> The Reset button may be a little subtle for people

i played around with a 2nd color option (matching the zoom controls) but scrapped it for some reason. can certainly set it back if that helps?

ideally it would also go below the zoom ctrls on desktop, but bottom-left is great on mobile and it seemed mess code-wise to adapt it (if there was no in/out animation, that would be different). will add to my wishlist notes though.  

> with the compass looking a bit like the Mapbox logo below 

it does but nowhere near as bad as DiCompass. i think that IS the MB logo haha. if you spot a better one let me know but i went with `TiCompass` because it has a light/dark:

image

PR #123, opened 2020-10-22
Feedback/contact form: open w/o new tab

### LINKED ISSUE #122: Feedback form: embed into site
...so that users don't have to leave in order to fill it out.

abettermap:
Closes #122

### DIFF
--- a/src/components/nav/Nav.tsx
+++ b/src/components/nav/Nav.tsx
@@ -16,6 +16,10 @@ import { AiOutlineQuestionCircle } from 'react-icons/ai'
 import { paths as routes } from 'components/config/routes'
 import { Settings } from './Settings'
 
+type NavProps = {
+  openFeedbackModal: () => void
+}
+
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     listLink: {
@@ -40,7 +44,8 @@ const useStyles = makeStyles((theme: Theme) =>
   })
 )
 
-export const Nav: FC = () => {
+export const Nav: FC<NavProps> = (props) => {
+  const { openFeedbackModal } = props
   const classes = useStyles()
   const { listLink, listItemIcon, smallerText } = classes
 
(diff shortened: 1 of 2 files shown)


rperlin-ela:
This does the trick. Can we change "Language Map Feedback" to "Contact and Feedback", matching what's in the sidebar? And should there be a "Report bug" option in the dropdown to match what's promised? I submitted some sample feedback which hopefully arrived and probably we should auto-generate a reply to feedbackers as well (which I haven't tried but I assume Google Forms can easily do). Or is it just easier for you to give me access to this Google Form, or for me to create a new one with an ELA account to make these changes? Thanks for putting this together and sorry for not being on top of this earlier!

abettermap:
Thanks for reviewing, Ross. I just pushed up a version that's a little more mobile-friendly and doesn't produce double scrollbars on desktop (in case you were seeing that). Same URL: https://deploy-preview-123--languagemapping.netlify.app/ If you were okay with it before then you're probably even more okay with it now 😄 

Re: your questions, Maya has access to the form so you two can hash that out. My role was just to make the embed happen, so unless you need anything else in that dept I'm going to merge.

rperlin-ela:
Ok all good!

fiddleHeads:
@rperlin-ela , I made the changes you suggested, and you can see and edit the form here: https://docs.google.com/forms/d/1CzZK-aKh0kyxR13DGEJ5t8S_mG9Cf2y-jwVviZ35fac/edit?usp=sharing. I just enabled email notifications, which will go to my email address because the contact form is located in my Google Drive. @abettermap is it possible to switch out the link for the form if Ross wants to host the form on his end so that he gets the email notifications? Thanks.

abettermap:
Yes, if you can't make him an owner/editor using the existing form let me know. Or it might be easiest if you set up an Outlook filter to auto-forward those emails to Ross and whomever else so that you can all see them.

rperlin-ela:
Great— thanks, Maya!

PR #124, opened 2020-10-22
Geolocation

### LINKED ISSUE #105: Geolocation
# Summary
Pan the map to user's current location.

### Can we just use the default example?
**If so, can skip most of the remaining stuff.** Only thing we might need to change is the button to make it match the others, or 

Default example: https://docs.mapbox.com/mapbox-gl-js/example/locate-user/

Available geolocation parameters [here](https://docs.mapbox.com/mapbox-gl-js/api/markers/#geolocatecontrol). **Near-future Jason:** it's also already a [React MB control](http://visgl.github.io/react-map-gl/docs/api-reference/geolocate-control).

### Button/control location and type?
1. Location: going to need more than just this control added to the UI (e.g. toggle census layer/s), so I'm thinking we could repurpose the geolocation pop-out into a catch-all for zoomy layery things.
1. Type: button or toggle?

### First-click behavior?
1. Page asks user for "Allow use of location"? (this is built into browsers)
1. User responds:
    1. "Block": show them a popup or similar indicating "welllll it's kinda hard to zoom to your location if we don't know what it is"
    1. Allow: map pans to current location

### Post-allow behavior?
1. What icon? This seems like a good standard-ish place to start from: https://docs.mapbox.com/mapbox-gl-js/example/add-image-animated/
1. How long should icon persist?

abettermap:

- Closes #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,20 @@ visible at[http://localhost:3000](http://localhost:3000).
 The page will automatically reload if you make edits. You will also see any lint
 errors in the console.
 
+#### SSL (https)
+
+If you prefer `https`, sans scary red warning in the browser, use a tool like
+`mkcert` and add these to your _.env_ file using its output:
+
+```bash
+HTTPS=true
+SSL_CRT_FILE=/The/Path/To/Your/Certificates/cert.pem
+SSL_KEY_FILE=/The/Path/To/Your/Certificates/key.pem
+```
+
+Steps may vary by OS but there are a lot of tutorials online. Just do a search
+for "create-react-app ssl" or similar.
+
 ### Test
 
 `yarn test` launches the test runner in the interactive watch mode. See the
--- a/src/components/map/GeocoderPopout.tsx
+++ b/src/components/map/GeocoderPopout.tsx
@@ -20,14 +20,28 @@ import { useWindowResize } from '../../utils'
 import * as utils from './utils'
 import * as MapTypes from './types'
 
+type PopoutContentProps = {
+  heading: string
+  explanation?: string
+}
+
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
-    popoverContent: { padding: '1em', width: 310 },
-    popoverContentText: { marginBottom: '.75em', fontSize: '0.8em' },
-    layersMenuPaper: { overflow: 'visible' },
+    popoverContent: {
+      '&:not(:first-of-type)': { marginTop: '0.5rem' },
+      '& > *': {
+        marginBottom: '0.3rem',
+        marginTop: '0.3rem',
+      },
+    },
+    explanation: {
+      color: theme.palette.text.secondary,
+      fontSize: '0.7em',
+    },
+    layersMenuPaper: { overflow: 'visible', padding: '1em', width: 310 },
+    // Toggle switches
     switchFormCtrlRoot: {
       marginLeft: 0,
-      marginTop: '0.8em',
     },
     smallerText: {
       fontSize: '0.8rem',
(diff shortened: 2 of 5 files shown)


abettermap:
@rperlin-ela @fiddleHeads here is the geolocation with a mere 10ish lines of code for the defaults. (I'm going to stop pasting in the deploy URL, I think you're familiar w/that, it's the last "Details" link above).

## Options to use?
You'll want to refer to the [available options](http://visgl.github.io/react-map-gl/docs/api-reference/geolocate-control#properties) and let me know if any should be changed or if they don't make sense to you. They're mostly all relevant but you can ignore these:

- the ones that start with `on` 
- style
- label
- auto

## My take

### 1. Options (above)
Personally I like it the way I have it (aside from style/layout stuff, although even the defaults for that aren't terrible- it does a little spinny animation while it's waiting for you to allow/block), but I'm only testing it on a simulated lat/lon so I could see how `trackUserLocation` could get annoying. BUT, disabling it could get confusing because that's not the standard in map apps. As in, you turn the thing on and the blue dot normally keeps moving with you. 

Note that on a non-mobile device, it shouldn't really move. It only uses your Wifi/IP to determine your location (mine was frighteningly accurate, like down to the _corner of my house_ accurate), not GPS (if equipped) + Cell + Wifi like phones.

Note also that `trackUserLocation` does NOT track and send the info anywhere. It's a misleading name, it just refers to the paragraph I wrote above this one.

### 2. Button location
Unless you anticipate the majority of users to be in NYC, and the majority of _those_ users to want to allow the usual:

image

...I probably wouldn't put the button front and center, visible 24/7 like it is. That's some pretty valuable real estate on mobile, so just depends on how high a priority it is.

#### 2.1 If you agree...

...where should I put it? It doesn't even need to be a button, it can probably be a toggle/switch thing. My vote would be inside the "Search by location", which would be renamed of course, but I think we can get away w/same icon for that btn.

#### 2.2 If you _disagree_...
...where should I put it? It should be a button and ideally match the style of the others, I'm just wondering where it would go. I have a couple ideas but I'll wait until you respond to 2.1 first.

## Testing
Ross, you're the man on the ground so you'll be the only true tester. Mask up and get out there! 😷  

Maya, if you want to know how to simulate location in Chrome try [this](https://developers.google.com/web/tools/chrome-devtools/device-mode/geolocation) and/or a non-Google article [here](https://medium.com/@williamkoehrsen/controlling-your-location-in-google-chrome-6c0b216d1ba1#:~:text=Manually%20Set%20Location%20Using%20Developer%20Tools&text=Find%20the%20%E2%80%9CMore%20Tools%E2%80%9D%20entry,%E2%80%9CSensors%E2%80%9D%20option%20as%20shown.&text=This%20will%20display%20the%20sensors,of%20the%20developer%20tools%20panel.)

## Disclaimers
If it's not already in the Help or About, might want to consider a "hey really tho, we swear we're not outside your window, mwahaha". (mwahaha optional but encouraged)

## Anything else?
Don't want to jinx it but this has been too easy so far. Am I missing anything? The intent is not analysis (e.g. how many arbitrarily approximated points are with 100m of me), it's just a convenience for users in the area. If so, once we settle on the stuff above, are we good?

abettermap:
f'real though, my laptop and i are about 10ft from the edge of the outer perimeter:

image

technology! 😱

abettermap:
fyi this is not working for me on mobile for some reason. was working-ish before so i'll look into it tomorrow.

in the meantime you can still try on desktop though.

abettermap:
scratch that, i had location disabled in both Safari and Chrome on my phone. derrrrp.

test away!

rperlin-ela:
I’m into it! Working great for me on mobile and desktop. Ironically, I’m going out of town till early next week so won’t be able to do much testing till then, but right now it’s looking great. Only quibble is that on desktop it zooms me way in on a few blocks. Not a big deal, but the zoom level on mobile is perfect.

All your points below make sense to me off the bat, as does making it a toggle in a renamed “Search by location”.

abettermap:
Sorry for delay, getting house ready for January weather in October. Single digits soon! 🥶 

Re: too far zoom on desktop, fine on mobile: not sure what either of those looks like on your devices (it might depend on a few things) but I just went with the same zoom as when you click a community (assuming you weren't already zoomed in close in that scenario, in which case it would keep the same zoom). 

Looks fine to me after this update but I'm simulating so let me know. You should be able to test even if you're not in NYC, you just won't see any points of course.

The control was easy to move, I just need some human wording and/or layout/UI suggestions for where I put it:

image
 
Bear in mind that we'll likely add a census toggle in there as well. If it gets too crowded or unruly, I have plenty of ideas, but haven't really had to mess with the layout yet.

### Other bells and whistles

I added some other ideas in this issue: #125

fiddleHeads:
Works great for me on desktop. I actually like the wording you have and the zoom to location inside the search by location dialog works fine, I think.

abettermap:
Cool, thanks for reviewing, I think we'll just sit tight on this until Ross is back in NYC for testing.

rperlin-ela:
Seems to be working pretty well for me now with consistent zoom level across platforms and browsers, so I think good to go.

Issue #98, opened 2020-09-26
Panels on map: miss u buddy 😞 


abettermap:
Small thing to some users perhaps, but I feel that the "panel floating on map" vs. what we settled on, "panel adjacent to map", sets map layouts apart. There are more non-floaters for a reason: it's a lot easier because you don't have to deal with any squirrelly offsets or padding to create the "perceived center".  

Old screenshot but you get the idea:
![image](https://user-images.githubusercontent.com/4974087/94326965-0e01cc00-ff65-11ea-823e-6be91bfdf778.png)

Technically had it "working" but there were just more and more scenarios that required the offset and it was very hard to keep up with (especially since the app has like 4 CSS breakpoints). Not a deal-breaker by any means, but it did feel more next-level to have a layout like that, not to mention inform users that there is more than just the initial area to explore:

![image](https://user-images.githubusercontent.com/4974087/94327012-823c6f80-ff65-11ea-9bdf-4948c9346d21.png)

PR #127, opened 2020-10-26
WIP: init "See related" functionality

### LINKED ISSUE #20: Add "See Related" functionality
# Summary
...from today's Zoom, in a nutshell:

**the elements in the Details panel should be clickable and will trigger the same filtering behavior as that of clicking a cell value in the table.** See "Original summary" section below for more details.

## UI considerations

1. How to make these elements appear link-like without a 💩 -show of overwhelming colors? The white and great look great as-is, as does the lack of underlines.

## Behavior questions

1. Define app-wide behavior/impacts .
1. Outline scenarios which might conflict with table filters.
1. ...and vice versa.

Ideally this will all be route-based so we can just sprinkle in route links as needed. "just" 😆 

## Code approaches

1. How to preserve/overwrite routes efficiently?
1. If going all-in with routes, will need to scrap at least rewrite the logic in `onRowClick` handler.
1. Full-circle back to [this](https://github.com/sindresorhus/query-string) I guess?

## Original summary

From Ross (paraphrased):

> “Show Nearby” would be amazing... I’m continually aware that in many cases we have large groupings of languages: several dozen Englishes, a dozen varieties of Spanish, same with Chinese, Arabic, and some others. I’m sure the way we display search results will bear the biggest burden here, but I’m hoping that we can open people’s eyes to the diversity within familiar "languages" like these that are actually “macrolanguages” or groupings. There won’t be a way just to go to "Chinese", for instance… I have no sense of the technical challenges, but it just occurred to me in terms of a dream “recommendation engine” that one variety of Chinese could lead people to others, for instance.

More detail is available in this [Sheets Data Schema thread](https://docs.google.com/spreadsheets/d/1X5Wqna2fFQzYnzqv6rKYGJr10byevr2oea6GTvq21w0/edit?disco=AAAAJs2_SUo), but basically the idea is to see features with similar attributes and encourage users to explore. This could be a linkable item inside a popup, for example (Jason):

> What if the popup showed “Top-level language: Chinese” (or whatever the attribute might be) as a link rather than just text?
>
> Not to get too far into routing here, but a URL can really reflect what it represents, so in the above case the link would point to something like “map.languagemapping.org/layers/languages/top-level-lang-family/Chinese”. And visiting the link would do several things: symbolize and/or label the layer by top-level language, filter by Chinese, and zoom to the extent of the results?

Just brainstorming for now, total wishlist and not in the SOW, but we can revisit later if time permits. It certainly ties in with the routing concept as well.

abettermap:

- Closes #20


abettermap:
@rperlin-ela total WIP but the deploy URL should be enough of a proof of concept to get the idea juices flowing.

Too much to cover here so if you're back in town Monday would you have time for a Zoom? I'll try to get a few big things out of the way here but let me know.

## Review

### The direction I've started with #20
I think we painted ourselves into a corner in the SOW a bit so I'd like to propose an alternative approach of "drilling down into what's already there" rather than "a second or third or fourth place to set filters". Table would still remain the sole source of filters, but this new approach still accomplishes goal of getting user to explore and see what's there. I feel it's more encouraging as it gives more of an idea of what's available than a potentially largely empty map with only a few dots.

### Specifically
The concepts of what I'm going for with things like:

1. https://deploy-preview-127--languagemapping.netlify.app/grid
1. https://deploy-preview-127--languagemapping.netlify.app/grid/Neighborhoods
1. https://deploy-preview-127--languagemapping.netlify.app/grid/Neighborhoods/Astoria

## Ignore

### Anything with panel headers

image
image

### Anything to do with state persistence
...e.g. you're like 2-3 levels deep in "grid" or "Sift" or whatever it should be called, then you click Details then back to Sift and it's not where you left it. This will be a challenge but I'll figure it out.

### Relationship of Sift to filters
Part of the sell for this is to show _all_ the available categories/columns 24/7 for each nested level. Some may need to be disabled (but not hidden) and/or have a count after them (e.g. `Ethiopia (0)`) and if for some reason user still clicks it, will end up at something like "No communities with Ethiopia. Click 'Clear filters' and you'll see the whole lot.

I tried to wrap my head around having 3-4 places to filter and if I can't do it as a user it's going to be really tough code wise. So hopefully the "see more like this" goal is still met (if not improved upon) by "see more like this regardless of filters".

### Probably other stuff
I just threw this together this weekend, hope it's an acceptable start towards the issue (and hopefully an alternative to "everything everywhere should be clickable even if it's a very-displayish-read-only thing like the legend".

## Talking pts for convo
- nav/drawer simplify
- is behavior of approach ok?
- wording: grid, sift, category. what do humans say?
- icons, ideas, category-level summaries, value-level summaries (e.g. a blurb on "Ghana" or "Harlem", independent of any specific community): Sanity CMS potential

## I'm loving it so far
Want to see all the places with Egypt? 💥  https://deploy-preview-127--languagemapping.netlify.app/grid/Countries/Egypt

abettermap:
oh btw i got the parse-by-comma working easily, e.g. "Turkey, Russia" gets split into each country for the Countries Sift, so the "unique values" count should be accurate (just need to change that wording tho as they're NOT unique). however, the logic for listing the communities is still based on verbatim. so, it's kind of the reverse problem- only "Turkey" (but not the combo with Russia) would be on this list: https://deploy-preview-127--languagemapping.netlify.app/grid/Countries/Turkey

probably another thing that i didn't make clear above is that the "See related" stuff can and will definitely still be sprinkled in various places throughout the app, it's just that it will be a lot more straightforward and intuitive if it doesn't affect the filters. the routing (URLs) is already set up so it's just a matter of throwing in links as needed. i'm going to argue against wrecking what we have in the UI/layout/style by making the dozens of items all clickable in Details and Legend though, there's certainly a better way, possibly a small menu trigger or something.

abettermap:
I just couldn't handle the boring side-by-side sidebar and map layout so among other things I started reinstating #98. I'll fight the map offset/center until the bitter end to make this work.

also with all the changes and additions coming, the two-item top-level nav has kinda run its course so i'm running with something more common and modern like 5-item bottom nav on desk:

image

 or the "X" close btn on mobile anymore, and got a smoke-and-mirrors version of the iOS "pill" drawer toggle.

image




also dropped the btn toggle btn, waste of space and far out of reach:

image

more details later

rperlin-ela:
Just grappling with all of it, great progress, lots to think about. Also want to share this with others at some point soon— not to have a lot of cooks in the kitchen, but because it's such a significant re-think. I'm not totally there, but I think I'm beginning to get a sense of what you're going for and how this way of doing things could meet the goal in a different way. I liked our button minimalism before, so I just want to be sure that what's under Explore feels sufficiently different from what's under Data, I'm a little confused about why I'd go there. I get (I think) that it's kind of a hub for functionality that will be sprinkled throughout the site (if not exactly making everything clickable then somehow still making sure there aren't too many dead ends). Happy to experiment with Sanity or anything else to add wording, summaries etc. for guidance.

abettermap:
> Also want to share this with others at some point soon— not to have a lot of
> cooks in the kitchen, but because it's such a significant re-think. 

Indeed, group feedback will be good once we're far enough along to get some input.

> I liked our button minimalism before

This is arguably as minimal if not more so than before. Previously there were 7 actionable things (including Help) at the top: 

image

Now there is one (the panel close btn) and 5 at the bottom, including the new one, without compromising the things that need to be clickable at the right times:

image

...which I also intend to add to the Categories view.

Same concept for Details (don't worry, I'll improve style on the "Clear selection"):

image

I probably suggested the persistent UI layout where there's always the Data/Clear/Clear btns at the top, but it's a bit of a space-waste to have the "Clear" stuff visible 24/7 everywhere, especially when there's no reason to clear selection from the Explore view, and no reason to have "Clear filters" shown when there are none. Arguably a user might want to clear filters from the Details view, but they're not super related to each other. If there's a strong need I can add it.

On the space note, keep in mind the incoming things: census stuff and, if approved, the "Categories" drill-down panel. So, more UI views means more navigation btns.

I tried to stick with these principles:

1. The things that are needed in all views can be triggered from all views in a single click (e.g. Help when you're in Details, or Data when you're in Explore)
2. Remove the things that aren't needed in a particular view (e.g. "Clear selection" isn't needed anywhere but Details)

> I just want to be sure that what's under Explore feels sufficiently different from what's under Data, I'm a little confused about why I'd go there. I get (I think) that it's kind of a hub for functionality that will be sprinkled throughout the site (if not exactly making everything clickable then somehow still making sure there aren't too many dead ends).

I think of it as a similar concept to the map: a way to browse the data without filtering it. From a "new user" perspective, especially one who isn't familiar with the data or project, Explore (the proposed Explore, that is, not the one in prod) is a gentler alternative to the table for diving into what's available. Instead of a wall of data with some potentially unfamiliar column names with no explanation (unless user clicks Help, of course), Explore lets you see what's there from a simpler, on-demand hierachical level. I think drill-down is an accurate term.

The table is really handy for finding things if you know what you're looking for but only so-so for browsing and exploring. As a user I want a way to see a list of all the Countries, for example, that are available before I make more decisions. This provides the opportunity to see what's out there without having to sort and click through 20 pages of data.

If we had autocomplete/select dropdowns instead of text fields, that would achieve the same thing, albeit in a duller scroll-heavy input format than a nice spacious grid or list view. But in that scenario, we'd be limited to that "what's available?" browsing in just the table rather than elsewhere in the UI.

One more +1 for the New Explore, and I believe I mentioned this above, is that you could arguably accomplish more see-related discovery _without_ filtering. Say I want to see a list of all the Western Asia communities but still leave the unfiltered dataset intact and see the full set in the map. Currently that's not possible since the only method, via table, would trim the set down by 1000 records. The user immediately loses all knowledge of the other communities unless they clear the filter.

Lastly, it's not possible to browse/explore data without obscuring the map. As a user I might want to see what's in the map while digging through what's in the data.

### Next steps

I'd like to get some basic "more like this" or "see related" functionality in place so you can see how it ties in with the New Explore. I feel that it's very much accomplishing the goal and I may just be explaining it poorly. Will keep you posted with updates and let's shoot for a Zoom tmrw if you're around.

rperlin-ela:
> Now there is one (the panel close btn) and 5 at the bottom, including the new one, without compromising the things that need to be clickable at the right times:
> 
> image
> 
> ...which I also intend to add to the Categories view.

Very nice.
 
> The table is really handy for finding things if you know what you're looking for but only so-so for browsing and exploring. As a user I want a way to see a list of all the Countries, for example, that are available before I make more decisions. This provides the opportunity to see what's out there without having to sort and click through 20 pages of data.
> 
> If we had autocomplete/select dropdowns instead of text fields, that would achieve the same thing, albeit in a duller scroll-heavy input format than a nice spacious grid or list view. But in that scenario, we'd be limited to that "what's available?" browsing in just the table rather than elsewhere in the UI.
> 
> One more +1 for the New Explore, and I believe I mentioned this above, is that you could arguably accomplish more see-related discovery _without_ filtering. Say I want to see a list of all the Western Asia communities but still leave the unfiltered dataset intact and see the full set in the map. Currently that's not possible since the only method, via table, would trim the set down by 1000 records. The user immediately loses all knowledge of the other communities unless they clear the filter.
> 
> Lastly, it's not possible to browse/explore data without obscuring the map. As a user I might want to see what's in the map while digging through what's in the data.

Ok, these are strong reasons. I guess I just have to see how the items within the grid will look.

> I'd like to get some basic "more like this" or "see related" functionality in place so you can see how it ties in with the New Explore. I feel that it's very much accomplishing the goal and I may just be explaining it poorly. Will keep you posted with updates and let's shoot for a Zoom tmrw if you're around.

Ok, I'm pretty much open most of the day tomorrow.

abettermap:
hey Ross sorry i didn't contact you today, had some stuff come up at home and also been fighting the new panel-on-map layout for the past two days but i think i got it. had to make a couple small concessions but not too bad. 

when you have a chance please test basically anything (same deploy URL) that involves zooming: zoom in/out/home (aka reset) btns, geocoder, boundaries, table filter changes, click community point, etc. I've tested on laptop, large monitor, and phone so far. iPad is forgotten again.

moving the Reset btn on desktop was an accident but i don't hate it, it's closer to the other controls now:

image

i scrapped the panels' Close btn on desktop, there were just too many scenarios to try and deal with, e.g. is panel open, how should it center, should the title move, should the map pan/zoom based on panel being open or closed, etc. etc. If that's cool then i'll leave it out but if you need it back and are okay with not having the map adapt to the open/close state of the panels on desktop then i can probably restore it without a lot of trouble.

some of those scenarios definitely occur on mobile as well but the Close btn (the pill bar thing) is more important on small screen due to limited space so i toughed it out on that one and left it intact, albeit the map still treats all the zoom/pan as if the panel is open, hopefully that's ok. i might have changed my mind on forcing open the panel to show Details when it's closed on mobile and a point is clicked, feels a little more intuitive to me but i'm open to leaving it.

i also scrapped a minor thing (being able to _close_ the panel by clicking a nav btn when panel is open). again, different set of headaches there, and as long as we have the Close btn on mobile i think it's sufficient for UI nav.

the only other hiccups ("only" haha, if you only knew!) are Boundaries click on mobile doesn't zoom to its full extent, slightly truncated, but i should be able to fix later. 

if you're wondering why the challenge and what i was after, it was basically this example: https://docs.mapbox.com/mapbox-gl-js/example/offset-vanishing-point-with-padding/

...which unfortunately does not work with any of our scenarios b/c that's vanilla MB and we're using a React wrapper around it with very different properties (sorry awful tech explanation). but that's the functionality i was going for and i believe it's  working now.

barring any major issues you see with this, or any testing problems you come across, i should be able to dive back into the "see related" tomorrow and get you a working demo by CoB.

abettermap:
Oh and one other nice thing that would not be ridiculously hard to add after today’s efforts is a “maximized” panel view on mobile. So basically if the user wanted to see any of the views like Details in a larger manner, more or less like the table or any of the modals, without the map taking up half the screen, that seems doable.

rperlin-ela:
Rapidly tested out the various things that involved zooming on iPhone, iPad, and desktop (Safari on Mac) and it seems mostly good, so with any luck you’ve mastered the old panel-on-map problems. Doesn’t bother me to have the panel open 24/7 on desktop. 

Only the iPad feels a little off— seems to be treated like a desktop in this respect (panel open 24/7), and the panel does take up a fair amount of space there, so ideally would either be accordion-able or shrink a bit. Don’t know if it’s related but sometimes on iPad the Reset (3D) button acts a little strange, putting a selected community off-map. I can’t reproduce it but relatedly I think the zoom home button combined with the Reset (3D) button was also acting a little strange, taking me to a kind of random slice of the map, definitely not the standard “home” view— though hitting zoom home a second time always got me to the right place. Sorry that that bit is vague and not reproducible.

With this deploy at the moment, there are also some issues with the table— left two buttons not working, table not going away after clicking “View results in map”— but I assume those are an artifact of all the work in progress

abettermap:
thanks for reviewing, and sorry i probably shoud've omitted ipad in the test request! i hadn't looked at it at all so hopefully it wasn't too much of a time-waste for you. will fix those issues though. 

table stuff is byproduct yes, easy fix.

with the extra room from the former "X" that closed the map btns, i'm going to squeeze the Reset btn in there too (sans text) instead of making it fly in all the time. it will be disabled until community is clicked, then disabled again once itself is clicked.

i believe i know the 3D issue you mentioned, will look into it.

going to work on these patches until noon MST then must get going on the See Related as i tentatively deadlined it for today. would at least like a proof of concept anyway. so, if you're available today, what's the latest you could Zoom?

abettermap:
Awesome. Let's set a time so I have a window to keep me focused. How's 7:30 EST work for you?

https://deploy-preview-127--languagemapping.netlify.app/

I fixed most of the issues you mentioned but wasn't 100% sure if we are talking about same Reset 3D issue, but it seems better now that I just close the popup in addition to resetting the pitch, on mobile.

I think the zoom-to-home or zoom-to-anything is off on mobile by a bit (vertically), which works in our favor because it centers the popup, but makes the other zoomy things a little off. I think I had some issues with that in the old layout as well, will need to look into it.

FYI the table is all routed up now so you can view the full details modal via URL, e.g. https://deploy-preview-127--languagemapping.netlify.app/table/75

In single-page apps, routing is used to show/hide views based on URL. This is handy because you can go back/fwd and keep your bearings with where you've been in the UI. But for some cases it takes a little thought, like closing the table (which has a route of `/table`).  I wasn't sure how to handle some of the scenarios so I just went with my best guess for table closing:

1. If previous route had "/table" in it (aka the full-modal Details view) then closing table goes to map. As does clicking the "View results in map".
1. If not, it goes "back".
1. ...but only if the history stack is not empty. In other words if you happened to share the `/table` URL and a user visited it directly, e.g. from an email, their history would be empty because that's their first page in that tab. So in that case, take them to the map on table close.

It's more straightforward then it sounds but will get tricky with nested routes like `/grid/Countries/Turkey` where I need to keep track of what the last view was in each panel. If you were in that route in the New Explore panel, for example, then clicked an item and were taken to Details, when you click Explore again you should see the list of Turkey links again, not the list of Available Categories, which is currently the case.

Anyhoo much easier over zoom so let me know how 7:30EST works for you. I'm open as well.

rperlin-ela:
Sure, 7:30 pm EST is fine by me, will take a look before that

abettermap:
sounds good. i just pushed some good progress, it should deploy soon. give it a try and check the commit log for highlights: https://github.com/Language-Mapping/language-map/pull/127/commits

really digging the lack of dead ends this presents to the UI!

rperlin-ela:
Cool, this looks very exciting and starts to bring it all together. No dead ends! 

I’m psyched about the chips. Think I agree with you about being consistent even if there’s just one instance of a particular category. I’m not 100% sure we need Status as a chip but at least (Primary?) Country would be great — assume given the elegance of this approach that should be a cinch to change. 

(Though this really nice setup makes it tempting— I know it would mean a schema change— to see about adding a kind of “Macro-Community” field so we could do some really cool grouping/tagging that doesn’t fit our fairly staid current categories, like all the Himalayan languages, Jewish languages, Indigenous...)

Clickable legend will be gold. I understand not wanting it all linky, but maybe something when you hover over? First inelegant stab at a blurb: “Click any category with an icon below to learn more”? Or "… to see all the relevant language communities”. Of course people may still try to click headings like Africa. My temptation, if it’s not too bad, is to include Size for the sake of consistency. And depending on how things work in the Explore panel (should the header there be something like “Explore all categories” instead of “Available categories”?), I think Size has got at least as much claim there as Status.

Yes, I think any Category except Language would use it as the heading for each card.

Can write if I have more thoughts/comments or happy to do a call if you think it would still be helpful, just send me a link... I’ll be at my computer till 8:30 EST anyway. Weekend’s a little busy but early next week all good too. Definitely seeing the merits of the approach you’ve taken and excited to see where it leads, but I guess a lot hinges on what happens in the Explore panel if in a sense that’s where we’re directing everyone.

abettermap:
## Talking points for zoom call
@rperlin-ela shoot me an email w/zoom link when you're ready. my camera isn't working so just audio/screenshare i guess.

### Categories (??) 
- approach ok?
- terminology first!
- avail cats ok?
- for top-level, card content?
- "Categories" or grid or sift or explore orrrrr
- anywhere else to sprinkle?
- grid/card layout for all? same heading/panel content (except for Details)?
- if so, what should heading/subheading/etc. be? summary too?
- how do filters play in? (will include "Clear filters" btn in each view if needed)

### Home
- linkable legend items
- keep for non world region?
- instr/wording

### Breadcrumbs
- config: icons, blurbs
- ideal: final item has legend swatch

### Details' see related
- wording: see related, more like this, ??
- chips vs drop-down vs ??
- lots going on. better overall UI options to handle The Busy? menu, toolbar, etc. Or maybe show/hide panel like Share and World Map? Should define top above-fold priorities first either way.
- how many see related's to show?
- what if only 1 instance? still link to cat?

abettermap:
emailed you a zoom link, hop on if you're free

rperlin-ela:
Two quick follow-up thoughts: 

After your fixes, I think I’ve narrowed down what’s happening on the iPad. Any time it’s in the 3D mode the reset button seems to take things too far down— with individual points it often means you can’t see it since you’re down below when you’re back in 2D.

Maybe it’s just the busy stackiness of the Details now, and potentially compounded by our taking out “Similar Communities”, but it occurred to me today that it’ll seem redundant to have Language, Region, and Country both immediately overhead and in the chips. Just putting it out there— I totally understand why you don’t just want all the existing stuff to be clicky, and the chips are nicely visually distinct from what’s above, but I could see this being an argument either against the chips or against the info above

abettermap:
noted on the ipad (phone too), on my list to fix. thanks for narrowing it down so well, definitely helps w/troubleshooting!

bugs aside, i've been wondering if the whole force 3D + reset btn is even the right approach. perhaps a toggle switch in Settings that says "Enable 3D zoom", turned on by default, and just drop the reset btn approach? Below the switch could be the subtle instruction of "Hold `Ctrl` or `Shift` or `CMD` or whatever + mouse drag to change the pitch at any time regardless of whether this setting is enabled" (obvi w/a shorter human-friendly version). Then the reset btn could be available 24/7.

or just drop the fly-to 3D entirely, although i really like it. would be better to just figure out the issue on ipad and down.

Re: Details, yes redundant and busy indeed, although i've spent all of 15 minutes on styling it after the new changes, so there's much room for improvement. i think you're onto something with a chips-only approach though, so let me see what i can do with that. i'm totally fine with clicky, it just has to _look_ clicky, without being overwhelmingly so, and not be a super-secret Super Mario Bros 3 fly above the castle ceiling option (loved that game!). i do not find myself clicking things like the legend items which look very similar to surrounding items (white text, etc.) but i think we have established that i am alone in this regard, so if the legend items are already encouraging clicks for `n - 1` users as-is, then i don't really need to restyle them.

i think chips could get overwhelming in the legend, and competitive for attention (even if they were gray), not to mention they occupy more vertical space, so most of the efforts to show the legend on first load to mobile users would be somewhat in vain.

let me see what i can do w/Details and we'll start from there.

rperlin-ela:
Don’t know if you need feedback at this point but these look all good to me, btw. Heads up that I’ll have a big deadline and have pretty limited time/ availability next week— but this week and weekend I’m happy to do whatever I can

abettermap:
ok got some things done in past couple days, much WIP yet but emphasis on the "P" as it's coming along well. 

got kinks to work out so won't elaborate but here ya go: https://deploy-preview-127--languagemapping.netlify.app/

one thing i finally fixed is returning to the view you'd expect to return to when you click the bottom nav btns or close the table. before on the former it would always take you to the top-level (aka landing, index, whatever) view (e.g. Details' "no community selected", or Explore's 6-card "Explore" view) rather than the last one you were at. this has become way more apparent with the nested views (e.g. Explore > Countries > Turkey) and it's a very very standard app/web-app pattern to return the user where they last were in each view/panel.

so, I THINK we're good on that.

haven't touched any of the ipad/phone offset issues yet, needed a break from that. this UI stuff is piece of cake compared to that!

anyhoo i think it's coming along and leading to non-deadends far more than i ever imagined the original click-to-filter approach i'm supposed to complete. hopefully all agree, will send email to group once test-friendly.

might be time to pay the Performance Piper soon, got some heavy things going on esp. with the 600+ Languages list rendering all at once. i'll ultimately set it up to work like the omnibox which actually only renders visible items as needed.

i've been told by one individual that the gradients are not good. i like them so feel free to settle the 2/3 vote on that! no feelings hurt, just got sick of the flat dark flatness and wanted to bring the 2000's back:

image

really digging the addl icons and reinforcement they provide to UI. obvi got some redundancy so i'll need a reminder on what to put in the footer for each card (the header, above title/flag, will indicate how many instances, filters taken into acct, out of total instances):

image

i'm thinking the cards at the final nested Explore view (the last one before you reach the sweet Details end-of-rainbow) should probably include Neighborhoods or Town, rather than glotto/iso, since that sets them apart from one another. we don't want this:

image

also aware of redundancy here, which needs endo as title:

image

like so:

image

IMO i'm not sure glotto/iso fits into these very well. i mean it _fits_ UI-wise:

image

...but i think until that final pre-Details view it should just be totals, like at this level (w/actual totals obviously):

image

**question for that view:** what should/could the footer be? probably examples/instances of that value? e.g. Caribbean would have "Kreyol, Garifuna..." etc.

## Details
let me know how you're feel about this section. for the ones that had _only_ a primary country and no 2nd, i just omitted that line from below the chips. if there is 2nd country, show em both:

image

btw if i haven't mentioned yet, and probably not mega-useful but kinda handy and easy to set up, is that the in-table Details is routed now, e.g. https://deploy-preview-127--languagemapping.netlify.app/table/3

works fine but the screwiness comes in with "Clear selection". i think that should probably just be omitted from that view entirely. on the other hand it's kinda nice to have the nuclear "F it, i'm out, take me to Details!". 

thoughts?

abettermap:
also aware of missing titles here

image

just winging all this so there will probably be other scenarios, will fix as i go

rperlin-ela:
> ok got some things done in past couple days, much WIP yet but emphasis on the "P" as it's coming along well.

Great stuff

> one thing i finally fixed is returning to the view you'd expect to return to when you click the bottom nav btns or close the table. 

Feels intuitive at first go

> i've been told by one individual that the gradients are not good. i like them so feel free to settle the 2/3 vote on that! no feelings hurt, just got sick of the flat dark flatness and wanted to bring the 2000's back:

Seems like a nice change to me, maybe I'm also getting nostalgic for those days, but let's get @markturin as the deciding vote

> really digging the addl icons and reinforcement they provide to UI. obvi got some redundancy so i'll need a reminder on what to put in the footer for each card (the header, above title/flag, will indicate how many instances, filters taken into acct, out of total instances):

Nice icons!

For the Explore categories, maybe I'm forgetting an issue, but should we just uniformly pluralize so that we have "Languages" etc to match the existing "Countries" and "Neighborhoods"? Everything except Status. I know the filtering means that in some cases we'll have just one or zero. It might be a placeholder as is, but I also wonder if there's a simpler way to indicate "15 Instances (Inc. Filts)"

I guess we now really have to think about how to present these potentially long lists of cards, especially Languages, Countries, Neighborhoods, Language Families. So first of all alphabetize, right? For regions, I guess as consistent as possible with the legend. I don't know if headers to break things up wreaks havoc, but as with regions in the legend I could see it being very useful with neighborhoods everything grouped under one of the five boroughs (Brooklyn, Bronx, Manhattan, Queens, Staten Island"). For languages, if the intention is to have endos at Home / Explore / Language as you showed, I just want to make sure people don't feel at sea and can tell it's alphabetized by the smaller Language.

Agree with you about the final nested Explore view. But this also depends on Categories only showing one instance of any given community, right? Now under Western Asia (for example) we see four instances of Classical Arabic etc (undifferentiated by neighborhood), which I feel like gums things up, and then jump straight to details from there. I understand you're trying to avoid too much circularity when someone could have just gone via Language, but...

> IMO i'm not sure glotto/iso fits into these very well. i mean it _fits_ UI-wise:

I'm totally down with taking ISO/Glotto out, feels too busy and wonky for this. 

> **question for that view:** what should/could the footer be? probably examples/instances of that value? e.g. Caribbean would have "Kreyol, Garifuna..." etc.

Sounds right

> let me know how you're feel about this section. for the ones that had _only_ a primary country and no 2nd, i just omitted that line from below the chips. if there is 2nd country, show em both:

Yeah, that's good, but when I said Primary Country only in chips I hadn't yet thought about redundancy issues here and I think we better just make the Secondary Country a chip too and totally eliminate the old country line, if that's ok. "Clear Selection" is better, but maybe to the right of "Share"? Not sure.

Any reason not to go the whole way and make Neighborhood a chip also? In that case I might push the Language name chip under the endonym and have the Neighborhood chip down to the left of country.

> btw if i haven't mentioned yet, and probably not mega-useful but kinda handy and easy to set up, is that the in-table Details is routed now, e.g. https://deploy-preview-127--languagemapping.netlify.app/table/3

Nice

> works fine but the screwiness comes in with "Clear selection". i think that should probably just be omitted from that view entirely. on the other hand it's kinda nice to have the nuclear "F it, i'm out, take me to Details!".

Yeah, I had the same feeling. I wouldn't expect to leave the table view, just to close the little window, so I'd take it out of this view.


Small issue, but after setting a filter, the omnibox looks a little funny, gets all scrunched up 

Screen Shot 2020-11-03 at 7 59 49 AM (2)

Another tiny one: "Soundview" has a comma after its name in the Neighborhoods category.

I noticed that "Explore/Language/Arabic%20(Moroccan)" takes in both "(Arabic) Moroccan" and "Judeo-Arabic (Moroccan)". On the one hand, this is kind of interesting, given our interest in doing more "See related" stuff, but I think for now we have to be consistent and just take in languages with the exact same name, right? Or is there another way?

rperlin-ela:
> (Though this really nice setup makes it tempting— I know it would mean a schema change— to see about adding a kind of “Macro-Community” field so we could do some really cool grouping/tagging that doesn’t fit our fairly staid current categories, like all the Himalayan languages, Jewish languages, Indigenous...) 

To try to get some great extra mileage out of the chips, would you object to giving this a try, in as simple a way as possible? I would add "Macro-Community" to the sheet/schema, not a mandatory column (i.e. not everything will have a value), there would be about a dozen different values. Maybe (certainly initially) no need it to appear at all in the Data table. In order to be a chip I guess it would have to be another category, but I'm hoping maybe no other impacts? Is there a way to try this without it being a huge lift for you?

markturin:
Morning Ross, and Jason, it’s been a long time since we were all in touch. I’m not sure that I’m fully up-to-date with what all this means, perhaps Ross (you and I) could have a zoom call soon so I can answer the questions and review anything that you think you need me to weigh in on? Many thanks to both of you for all of your work from an extremely busy time in the term, thinking of you and all our friends in the states today, Mark

rperlin-ela:
Thanks, Mark, a big day... Sorry for the tag in medias res, more a note to me and Jason, but I think there will be something to look at very soon and I’ll certainly be game for a zoom call between us if that seems useful.

abettermap:
Thanks Ross, all good points. I have to be careful not to open too many worm-cans here or it will lead to additional days/weeks/months/decades (🤔 ) of work! so I'll need to stick with the "choose our battles" approach on some of them.

> For the Explore categories, maybe I'm forgetting an issue, but should we just uniformly pluralize so that we have "Languages" etc to match the existing "Countries" and "Neighborhoods"? Everything except Status. I know the filtering means that in some cases we'll have just one or zero. 

Makes sense but the further i deviate from the data-driven approach the more complicated and error-prone it gets. At the top level in Explore I could hack something together if that is a high-priority, but then it would be inconsistent from there downward in the hierarchy. But since everything is tied very tightly to the verbatim column names, it would mean a lot of checks and customizations in a lot of places just to pluralize it. I mentioned the other day how a CMS would come in handy for little trivialities like this but sadly we don't have one. :(

> It might be a placeholder as is, but I also wonder if there's a simpler way to indicate "15 Instances (Inc. Filts)"

yes, placeholder, and yes probably a simpler way. i'm not sure how useful showing the original instance count in addition to the filtered count will be to the user. i think all they will care about, and maybe not even, is how many instances there are and an indication that filters are causing the total to deviate from the original total. so i'm thinking "15 items" and if there were originally more than that and filters are affecting it, then a small filter icon (or continue with our orange dot) next to it.

honestly that's getting pretty complex though and may lead to more confusion than it's worth. so i'm leaning towards (for the mid- and final-level nesting depth) just showing the total at the top of the category panel rather than within the individual card items. so, user would see the total after they've clicked on a card rather than in the card.

also as a supplement to the filtered totals, i intend to include the filter indicator dot (or icon) at the top of the panels whenever needed and reuse the same blurb as the one below omnibox.

please note that the filters indication challenge is not unique to our "table as the sole source of filters" setup. it's inherent to any UI where there are filters affecting the results regardless of how/where the filters get set. typically the filters are shown and set in an obvious, dedicated, easily-or-always visible spot, but having the table set the filters makes that a little wonky. if we didn't use the table as a way to set filters, there could be a dedicated form/modal elsewhere. this was my original plan but then switched to the table since it's already built in. but now that the table filter fondness levels are so high, there is not a lot of room for other intuitive app-wide filtering options. if there was openness to scrapping the table filters, or using them as a means to only filter and sort what's in the table (which was my original approach but it led to confusion) without making it confusing, then perhaps an app-wide-accessible filter form would be an option. not in this version, but SOW.3 maybe. :) 

that said, i think our subtle dot and blurb sprinkled into the panels will make it hit home a bit more, and i'll be sure to include links to the table and "Clear filters" in said blurb. 

> I guess we now really have to think about how to present these potentially long lists of cards, especially Languages, Countries, Neighborhoods, Language Families.

yes they're a bit unruly. i've got some ideas. i think a nav list at the top with `A B C D E F` etc., each linking to the to first instance (and yes i'll have sorted items) of the item starting with that letter, would be a great start. i'd have to either make the a-z list "sticky" to the top, or add a "Back to top" btn. the tool i'm going to use to make the lists performant has an [example](https://codesandbox.io/embed/github/bvaughn/react-window/tree/master/website/sandboxes/scrolling-to-a-list-item?fontsize=14&hidenavigation=1&theme=dark&view=preview) similar to the jump-to, albeit sans letters.

> So first of all alphabetize, right? 

sure, can do.

> For regions, I guess as consistent as possible with the legend. I don't know if headers to break things up wreaks havoc

it does. on that same note of deviating away from data-driven UI, sans CMS, the potential for errors and inconsistencies is quite high and the code is difficult to maintain. the World Regions continent config was all hand-rolled on my part since there is no tie to the data whatsoever. it may look very simple on your end but not having anything to work with data-wise on mine is a dicey endeavor. CMS could solve this and then some, but much time required for setting that up.

> but as with regions in the legend I could see it being very useful with neighborhoods everything grouped under one of the five boroughs (Brooklyn, Bronx, Manhattan, Queens, Staten Island").

agreed, def a good idea, but i'm going with a hard No on that one. more deviation and config and potential problems. most of the stuff i've added has been quite a bonus, so i need to be careful on how far to take the wishlist requests. you should certainly make an issue for it though as it would be a good way to organize the neighborhoods!

> For languages, if the intention is to have endos at Home / Explore / Language as you showed, I just want to make sure people don't feel at sea and can tell it's alphabetized by the smaller Language.

yes good point. so would the Language be the primary text for the card at that level then? and endo the "intro" text? i think that would be fine, or maybe instead if we had Language as the intro, Endo as the primary (middle, second item of card), and examples as the footer, seeing Language first might prevent the at-seaness.

> Agree with you about the final nested Explore view. But this also depends on Categories only showing one instance of any given community, right? Now under Western Asia (for example) we see four instances of Classical Arabic etc (undifferentiated by neighborhood), which I feel like gums things up, and then jump straight to details from there. I understand you're trying to avoid too much circularity when someone could have just gone via Language, but...

the pre-Details level will always contain the same hierarcy if that's what you're asking: `Explore > Column (w/instances list) > List of instances of Column`. as for "undifferentiated by neighborhood", i'll fix that. each card should be unique in each view. if that's the only issue here then disregard my first sentence. 

> I'm totally down with taking ISO/Glotto out, feels too busy and wonky for this.

consider it done.

> Yeah, that's good, but when I said Primary Country only in chips I hadn't yet thought about redundancy issues here and I think we better just make the Secondary Country a chip too and totally eliminate the old country line, if that's ok. 

No problem.

> "Clear Selection" is better, but maybe to the right of "Share"? Not sure.

hmm disagree here. i feel like right-side = least important or least used (no offense to Share, you're great buddy). i don't feel like it belongs in the Media line either, but if we have a stacking-averse policy then i don't know where to put it. i'm open to other ideas, but if we don't have any then let me know if you'd prefer same line as Media or its own line. IMO i wouldn't mind it below the gray line above the Description. if it was more obvious to users that clicking empty space in the map has (or at least used to have before i broke it) the same affect, we wouldn't need a button at all. additionally i could make the "X" in the popup do the same thing. 

i kind of lke the Clear selection btn tho so let me know what you think.

> Worth going the whole way and making Neighborhoods into chips also? I know that would means a lot of chips in some cases...

yeah that's getting pretty chip-heavy, and instances like [this](https://deploy-preview-127--languagemapping.netlify.app/details/55) or [this](https://deploy-preview-127--languagemapping.netlify.app/details/887) would be nutso. aesthetically i quite like Neighborhoods as it is now but i understand the need for consistency and clickability, so your call. if we went all-in with the chips (take the gambling pun alllll the way) then we'd have a rather dull Title/Chips/Links intro. functionality and goals are more important tho so your call. i guess the silver lining would be more room for a Clear Selection btn on its own dedicated line. 

it doesn't look like there are THAT many records with tons of Neighborhoods AND two countries so maybe i'm being paranoid about the over-chipping. i'll try it and see how it plays out.

> works fine but the screwiness comes in with "Clear selection". i think that should probably just be omitted from that view entirely. on the other hand it's kinda nice to have the nuclear "F it, i'm out, take me to Details!".
> 
> Yeah, I had the same feeling. I wouldn't expect to leave the table view, just to close the little window, so I'd take it out of this view.

will do

> Small issue, but after setting a filter, the omnibox looks a little funny, gets all scrunched up

yuck! will fix.

> Another tiny one: "Soundview" has a comma after its name in the Neighborhoods category.

data issue on your end. record 823 has a trailing comma.

> I noticed that "Explore/Language/Arabic%20(Moroccan)" takes in both "(Arabic) Moroccan" and "Judeo-Arabic (Moroccan)". On the one hand, this is kind of interesting, given our interest in doing more "See related" stuff, but I think for now we have to be consistent and just take in languages with the exact same name, right? Or is there another way?

ah, good catch. consistency is intended, this is just a byproduct of me changing the logic to "value includes" from "value matches exactly" so that the parsed Neighborhoods and Countries would be matched non-verbatim. i think i can just check for the presence of comma before using "value includes". suffice to say we can never have commas if the intent is not a list of values, won't work right at all.

> Don't have an immediate solution but I do wonder, nav-wise, if it's confusing to have "Home" to be a place where you search and the magnifying glass associated with search take you somewhere without a search bar.

agreed, i noticed as well. all the names are pretty arbitary but i think Home and a Home icon is too standard to pass up. i'm sure React Icons misses you, maybe you could peruse for an Explore alternative. :smiley: i think binoculars stand out to me, but there may be others.

> To try to get some great extra mileage out of the chips, would you object to giving this a try, in as simple a way as possible? I would add "Macro-Community" to the sheet/schema, not a mandatory column (i.e. not everything will have a value), there would be about a dozen different values. Maybe (certainly initially) no need it to appear at all in the Data table. In order to be a chip I guess it would have to be another category, but I'm hoping maybe no other impacts? Is there a way to try this without it being a huge lift for you?

if it's the exact same functionality as the others and it doesn't involve any joins or crazy config stuff or any other special scenarios, then shouldn't be a heavy lift. you should definitely spearhead the React Icons movement on that one tho!

will that column be used directly for the **Census stuff** as well?

rperlin-ela:
> Thanks Ross, all good points. I have to be careful not to open too many worm-cans here or it will lead to additional days/weeks/months/decades (🤔 ) of work! so I'll need to stick with the "choose our battles" approach on some of them.

Understood — just going to try to get you some quick responses here and then let you get back to it.

> Makes sense but the further i deviate from the data-driven approach the more complicated and error-prone it gets. At the top level in Explore I could hack something together if that is a high-priority, but then it would be inconsistent from there downward in the hierarchy. But since everything is tied very tightly to the verbatim column names, it would mean a lot of checks and customizations in a lot of places just to pluralize it. I mentioned the other day how a CMS would come in handy for little trivialities like this but sadly we don't have one. :(

Ok, gotcha. On second thought, what if we just change "Countries" and "Neighborhoods" to their singular equivalent in the spreadsheet/schema? I know that's also a fair bit of work for you, but now that I look at it, maybe that change across the board wouldn't be a big deal and would keep things consistent. Yes, we allow multiple values in those, but that will speak for itself.

> honestly that's getting pretty complex though and may lead to more confusion than it's worth. so i'm leaning towards (for the mid- and final-level nesting depth) just showing the total at the top of the category panel rather than within the individual card items. so, user would see the total after they've clicked on a card rather than in the card.
> 
> also as a supplement to the filtered totals, i intend to include the filter indicator dot (or icon) at the top of the panels whenever needed and reuse the same blurb as the one below omnibox.

Sounds good, not a top priority on my end, so go for now with what you think best.

> yes good point. so would the Language be the primary text for the card at that level then? and endo the "intro" text? i think that would be fine, or maybe instead if we had Language as the intro, Endo as the primary (middle, second item of card), and examples as the footer, seeing Language first might prevent the at-seaness.

Yeah, I did kinda like leading with the endonym, in a way that echoes Details, and ideally it wouldn't be toooo small under Language, but this seems like the most straightforward way to go

> the pre-Details level will always contain the same hierarcy if that's what you're asking: `Explore > Column (w/instances list) > List of instances of Column`. as for "undifferentiated by neighborhood", i'll fix that. each card should be unique in each view. if that's the only issue here then disregard my first sentence.

Ok, at least differentiating by neighborhood is essential and will probably make things fine. Otherwise I was just thinking, and maybe it's too convoluted, that I just like how /Explore/Language leads to one card for each language and only then can you choose from the multiple communities (e.g. Explore/Language/Arabic%20(Classical)), while all the other categories show all the communities together, like here

Screen Shot 2020-11-03 at 2 52 50 PM (2)

But not a huge deal, I think, if the status quo is really more consistent

> i kind of lke the Clear selection btn tho so let me know what you think.

ok, go with your instinct and leave as is I still think it's pretty solid.

> > Worth going the whole way and making Neighborhoods into chips also? I know that would means a lot of chips in some cases...
> 
> yeah that's getting pretty chip-heavy, and instances like [this](https://deploy-preview-127--languagemapping.netlify.app/details/55) or [this](https://deploy-preview-127--languagemapping.netlify.app/details/887) would be nutso. aesthetically i quite like Neighborhoods as it is now but i understand the need for consistency and clickability, so your call. if we went all-in with the chips (take the gambling pun alllll the way) then we'd have a rather dull Title/Chips/Links intro. functionality and goals are more important tho so your call. i guess the silver lining would be more room for a Clear Selection btn on its own dedicated line.
> 
> it doesn't look like there are THAT many records with tons of Neighborhoods AND two countries so maybe i'm being paranoid about the over-chipping. i'll try it and see how it plays out.

Great, let's give it a try. Neighborhoods, even if chips, could still be on their own line, right?

> data issue on your end. record 823 has a trailing comma.

My bad! We're slowly working on that QC...

> agreed, i noticed as well. all the names are pretty arbitary but i think Home and a Home icon is too standard to pass up. i'm sure React Icons misses you, maybe you could peruse for an Explore alternative. 😃 i think binoculars stand out to me, but there may be others.

Agreed, let's go with binoculars.

> if it's the exact same functionality as the others and it doesn't involve any joins or crazy config stuff or any other special scenarios, then shouldn't be a heavy lift. you should definitely spearhead the React Icons movement on that one tho!

Ok, much appreciated. If it's ok with you, I can upload a new tileset in an hour or two with a "Macro-Community" column to test this out. I think if this works, even very unobtrusively, it will bolster the argument for the power of Explore.

Icon-wise, the most obvious thing is a group of people like GrGroup or IoIosPeople, and it seems to me distant and different enough from the "Community" icon under Status. A more abstract image of unity or merger could also work, but I haven't seen quite the right one yet.
 
> will that column be used directly for the **Census stuff** as well?

No, I don't see this new "Macro-Community" column having anything to do with the census.

abettermap:
> Ok, gotcha. On second thought, what if we just change "Countries" and "Neighborhoods" to their singular equivalent in the spreadsheet/schema? I know that's also a fair bit of work for you, but now that I look at it, maybe that change across the board wouldn't be a big deal and would keep things consistent. Yes, we allow multiple values in those, but that will speak for itself.

yeah doable, shouldn't be terrible. mainly a very-careful find/replace on my end. you should _add the two new columns but leave the existing ones intact_ for the time being, then upload. this will keep the dev and prod working without pulling the rug. let me know when you've uploaded and i'll make the switch.

also please update the Data Schema sheet accordingly when you have a moment. i know we don't refer to it often but it's still technically the The Source of Truth.

> Ok, at least differentiating by neighborhood is essential and will probably make things fine. Otherwise I was just thinking, and maybe it's too convoluted, that I just like how /Explore/Language leads to one card for each language and only then can you choose from the multiple communities (e.g. Explore/Language/Arabic%20(Classical)),

i think what you're asking is 1-2 more levels of depth maybe? like `Explore > Status > Liturgical > Arabic (Classical)`, where that last endpoint is a list of the Liturgical Arabic languages? yes that is throwing off the simplicity but i think i understand the idea. 

if i understand correctly then all of the variations of `Explore > Column > Column Value > Language Name` except when `Column = Language`, correct? if that's the case then only `Explore > Language > Language Name` would stop at that depth and all the others would follow the same pattern?

tough one to explain haha.

> Great, let's give it a try. Neighborhoods, even if chips, could still be on their own line, right?

sure. i think it will look weird since all the chips are so similar tho. might need a different color background for the Neighborhood chips, just to break up the wall of red chips.

> Ok, much appreciated. If it's ok with you, I can upload a new tileset in an hour or two with a "Macro-Community" column to test this out. I think if this works, even very unobtrusively, it will bolster the argument for the power of Explore.

sure thing. let me know when uploaded. 

> Icon-wise, the most obvious thing is a group of people like GrGroup or IoIosPeople, and it seems to me distant and different enough from the "Community" icon under Status. A more abstract image of unity or merger could also work, but I haven't seen quite the right one yet.

there's something wrong with `GrGroup`, i couldn't change the color, so went w/the ios one.

binocs look great btw.

rperlin-ela:
Good all around. New tileset is in there. Data schema updated. Yes, I think you've got it about the levels of depth, hopefully not too painful and we can revisit if it doesn't make sense.

abettermap:
nice, i'll start working the new column into the mix once i get some cleanup done. need to be able to reuse the same component over and over rather than copying/pasting like i've got it now.

fyi the totals on the individual cards- to obtain those for the mid-level nested views, unless i've approached it wrong the code has to loop over each row in the dataset to find all the unique values for as many instances there are in that view. e.g. 180ish countries X 1000+ records, just to get a total. it's also getting the values for the "examples" in the card footer, but technically we'd only need the first 5 unless it had to be sorted first.

does it matter which 5 are used, as long as it's sorted? like could it be any 5 South America Languages providing they're in order? another hard one to explain 🤷‍♂️

abettermap:
also going to hold off on the Level 5 depth for now. got about 1k things to get working first and since this approach and UI decisions are literally happening by the comment, still with the looming uncertainty that this is not even what was asked for or desired, i have to think it through before i move into Level 5. in theory it will just be reusing the same code with different logic, but let's see how the first 3 levels work out first.

rperlin-ela:
Sounds good. As soon as you’re ready you to share (or want me to share) with Mark, Maya, Dan, let’s go for it

I don’t think it matters which languages are used as long as they’re in order.

abettermap:
Will do. Really struggling with the code logic to make all the scenarios work. Hopefully will defeat it today and work on the schema update and chips tomorrow.

abettermap:
https://deploy-preview-127--languagemapping.netlify.app/

i THINK i've got the scenarios down now, including the Level 5 (e.g. `/Explore/Countries/Belgium/Dutch`) but i've got about 1.5 days into it and am very much burned out on the immensely convoluted code to make this happen, and still trying to grasp how click-to-filter is a potentially preferable alternative to this when it would lead very quickly to unintuitive near-dead-ends and is not conducive to discovering more language communities, which i thought was the goal so i'm kind of at a loss as to what is needed.

please test all the scenarios for functionality. i'm sure there are 1k design things to fix and yes i'm aware that there are placeholders like "COUNT" etc., so that will have to stay on the todo list for a minute, as will dealing with the long lists. 

didn't have time to do schema or Details updates, been grappling with the dozens of scenarios involved in this setup, so i'll have a walk in the park to look forward to tomorrow when compared to today, which was possibly the most complex code logic i've had on this project. 🏆  what doesn't kill ya, right?

anyhoo make sure to also test the scenarios with comma'ed things, and the `/Explore/Language/Something` level since that's what threw off most of the consistency. also test the Level 4's comma'ed instances where one of the values does not occur anywhere in the dataset as the first (primary) value for any record. Afghanistan is what I've been testing and it seems to be working properly: https://deploy-preview-127--languagemapping.netlify.app/Explore/Countries/Afghanistan

kinda hard to gauge it though without doing some math in the table. that route for example has 8 items:

image

...but there are 12 Afghanistan records in the table. i believe it is due to the Country-Language unique combination count in the panels (e.g. Pashto and Dari each have 2 instances), which is intended and correct. whew! 😅

abettermap:
btw just knocked out the schema change in like 10 minutes (haven't pushed yet). nice work adding that, let's go ahead and keep the old redundant columns until this branch is merged.

abettermap:
also are you sure you want the hyphen in the Macro-Community? aside from ISO it's the only one:

image

easy find/replace on my end, LMK

abettermap:
adding macro and including each country was piece of cake but i think it's getting pretty busy and could get out of control fast if neighborhood gets chipped:

image

open to other options here, i know it's important to be able to click, just don't want to over-chip it you know?

abettermap:
bit less chaotic w/o the red. got plenty of color w/in the icons, can get away with more busy if using gray chips:

image

my take is that currently it looks very good style-wise having neighborhoods (or town) on its own line and different from the others, unchipped:

image

it's not just a spacing thing, it just is broken up well with the variations. but we need clickable, so two proposals come to mind:

1. leave it as is, but also chip just the first neighborhood
1. replace the `, ` (on my end. don't touch those commas!) with ` | ` and make each clickable.

i like option 2:

image

it doesn't encourage as much clickage but then again i feel the same way about the legend items so i'm probably not the best judge. if that's true though, i think given the approach to dealing w/multiple neighborhoods makes it more acceptable to drop them down a clickage notch. if it helps, i can throwin a subtle dotted underline or something, and/or use the "local" icon next to each one so its slightly more chippy, along w/a subtle outline to increase chippiness (the pipe ` | ` would be unneeded then). this would be more consistent w/the icon-per-clickable-item chips approach.

thoughts?

i pushed up a non-clickable piped version, along w/a bunch of mobile and other UI tweaks. hasta mañana.

rperlin-ela:
> i THINK i've got the scenarios down now, including the Level 5 (e.g. `/Explore/Countries/Belgium/Dutch`) but i've got about 1.5 days into it and am very much burned out on the immensely convoluted code to make this happen, and still trying to grasp how click-to-filter is a potentially preferable alternative to this when it would lead very quickly to unintuitive near-dead-ends and is not conducive to discovering more language communities, which i thought was the goal so i'm kind of at a loss as to what is needed.

Yeah I understand, and I can imagine the convolution. I think this is all for the good— for people coming from Details and clicking on chips, it'll definitely mean more discovery. For people coming straight to Explore, it'll be more drilldown for people who don't have something to search for right off the bat in the omnibox.  But I feel we could use some fresh eyes, so it might make sense just to do some minimal cleanup and then take this to the team. 

> anyhoo make sure to also test the scenarios with comma'ed things, and the `/Explore/Language/Something` level since that's what threw off most of the consistency. also test the Level 4's comma'ed instances where one of the values does not occur anywhere in the dataset as the first (primary) value for any record. Afghanistan is what I've been testing and it seems to be working properly: https://deploy-preview-127--languagemapping.netlify.app/Explore/Countries/Afghanistan

By comma'd things, you mean Neighborhood and Country, right? 8 languages for Afghanistan in Explore feels just right and would match what we'd have in the table if we could just filter that value solo without Iran, Pakistan, Tajikistan, Turkmenistan— so all good there. What feels a little a funny is having 5 items under Country when you've presumably filtered just to see one country, but I guess that's an inevitable artifact of our shmooshing primary and secondary country together. 

Likewise with the additional neighborhoods— this is probably more my problem to think through, but now that Additional Neighborhoods are getting cards that you can click on, people may be surprised not to be going exactly where they thought they'd be going. Maybe the primary neighborhood can somehow be highlighted or emphasized a little in Details? And I know I wanted to alphabetize everything, but would it make sense/be possible to make an exception for /Explore/Language/Something in a way that mirrors Details, listing Neighborhoods in the primary/secondary order we have in the spreadsheet? I know that multiple "Primary + Secondary" combos get dropped in together here, the order in which they're dropped in probably doesn't matter but again I'm thinking that it would be good to find some way to highlight/emphasize the cards for Primaries.

Btw, all non-NYC communities are getting dumped into every NYC neighborhood, e.g. Explore/Breezy%20Point/, which should just have Irish English but seems to have Irish English plus all 97 non-NYC's.

Excellent that we can clear or view the filter now from anywhere in Explore. For clearing, does the user need to click away or refresh or something? I wonder if there should also be a note, like right after "You can clear or view them if needed", saying "You can also set a filter to view a specific set of communities on the map". Minor note that the yellow dot is getting stretched.

Screen Shot 2020-11-05 at 10 27 17 AM (2)

> bit less chaotic w/o the red. got plenty of color w/in the icons, can get away with more busy if using gray chips:
> 
> image

The gray is good!

> so two proposals come to mind:
> 1. leave it as is, but also chip just the first neighborhood
> 2. replace the `, ` (on my end. don't touch those commas!) with `|` and make each clickable.
> 
> i like option 2:

Ok, I hear all your arguments, let's go with 2!

> also are you sure you want the hyphen in the Macro-Community? aside from ISO it's the only one:
> 
> image
> 
> easy find/replace on my end, LMK

Sure, let's make it "Macrocommunity". I'll upload another tileset with the fix. 

Great that you were able to get this in there, seeing those chips looks good. Hope it's not too bad to finish setting up Macro so it shows up under Explore and filters right. I assume it's getting 504 for all of them right now because that's the number that **don't** have macrocommunities!

> yes i'm aware that there are placeholders like "COUNT" etc., so that will have to stay on the todo list for a minute, as will dealing with the long lists.

No worires, though I feel like all this will go a long way towards making it all feel helpful and interesting rather than duplicative and confusing

abettermap:
> But I feel we could use some fresh eyes soon, so it might make sense just to do some minimal fixes and then take this to the team asap.

feel free to share whenever you want. i'm still trying to catch up with the fixes, let alone finish the functionality, so your call when you want to share. opening the door to more feedback means infinite potential for suggestions and criticisms, so it's challenging to find a balance between "check it out now but ignore this and this and this and that and this since they're not ready" and "wait until i finish everything so there's no confusion about what's ready".

> By comma'd things, you mean Neighborhood and Country, right?

yes

> What feels a little a funny is having 5 items under Country when you've presumably filtered just to see one country, but I guess that's an inevitable artifact of our shmooshing primary and secondary country together.

yes. a non-1:1 relationship is a tricky concept.

> Likewise with the additional neighborhoods— this is probably more my problem to think through

yes. this seems like it's been a challenge since Day 0. tough question to answer on your end indeed, but w/o a relational database or breaking out primary/secondary/nth-ary into more records, i'm at a loss on how else to handle it. way too late in the game for that anyway! 

> Maybe the primary neighborhood can somehow be highlighted or emphasized a little in Details? 

for sure. i think i'll go with non-chipped, piped text, and make the first one white, with the local icon, and the rest off-white. all clickable, so no dead-ends.

> And I know I wanted to alphabetize everything, but would it make sense/be possible to make an exception for /Explore/Language/Something in a way that mirrors Details, listing Neighborhoods in the primary/secondary order we have in the spreadsheet? I know that multiple "Primary + Secondary" combos get dropped in together here, the order in which they're dropped in probably doesn't matter, but again here too I'm thinking that it would be good to find some way to highlight/emphasize the cards for Primaries.

having a hard time wrapping my head around the suggestion and a harder time thinking how to code it. if you want, i can just not sort them and they should end up in the same order? since it's a consolidated set of uniques though, from different communities, i'm not sure how we'd know which are primary vs. secondary since a language might be primary for some but secondary for others. 

unless i'm misunderstaning the complexity, you'll have to settle for unsorted or sorted in this round.

> Btw, all non-NYC communities are getting dumped into every NYC neighborhood, e.g. Explore/Breezy%20Point/, which should just have Irish English but seems to have Irish English plus all 97 non-NYC's.

ok will look into it.

> Minor note that the yellow dot is getting stretched.

gahhh. what browser? i'm not seeing it in Chrome or Safari desktop or ios. 

> Sure, let's make it "Macrocommunity". I'll upload another tileset with the fix.

cool let me know when it's up. don't need the redundant column like w/Country and Neighborhood since this column is not in prod.

> Hope it's not too bad to finish setting up Macro so it shows up under Explore and filters right. I assume it's getting 504 for all of them right now because that's the number that don't have macrocommunities!

yep no problem, will do.

> No worires, though I feel like all this will go a long way towards making it all feel helpful and interesting rather than duplicative and confusing

like my dad says, "all we can do is the best we can". so in a sea of complex tasks and more on the way, an obvious placeholder was my best and least confusing option for the interim, and it's obviously not going into prod. if i remove then it's empty space, inconsistent w/the others. if COUNT is confusing then i will just use the word PLACEHOLDER or something.

abettermap:
it probably doesn't get TOO much more chaotic than this, so think this will work for you?

image

the secondary hoods are clickable.

how do you want to handle Town-only non-NYCs? i could go either way. one on hand, consistency w/in Details. on the other, `/Explore/Town` is not really accessible or advertised. not to mention i would have to deal with the commas in a different way this time, because the logic is set up to split any value containing `, ` into parts. so it's not digging the commas here:

image

`Town` also doesn't have its own config, so things like the panel title icon and card icons would be inconsistent or i'd have to hack something together to make it use Neighborhood's icon.

i guess the easiest approach is just to not make it linkable. i can style it as needed, but non-linkable would be a third style after Primary neighborhood and secondary's. so, getting a little more complicated code-wise and inconsistent UI-wise.

thoughts?

rperlin-ela:
﻿Chinese looks good. Non-linkable for towns sounds fine. Stretched yellow dot was Safari desktop, big monitor, but also seeing it on iPad. Not sorting neighborhoods seems worth a try if it’s likely to replicate the spreadsheet— understood if there’s no way to style the primaries because it’s a consolidated set

abettermap:
i can't reproduce the stretched dot thing, maybe i've fixed it locally but if not then i might need you to Inspect it in Safari and screenshot the HTML.

this is unsorted version of `/Explore/Country/Ghana/Akan`

image

if that's better than unsorted i'll leave it. i think Option C you're asking for is "find out if it's a primary neighborhood for _any_ of the instances then sort by those followed by secondary". if so then that will have to be wishlist as i absolutely cannot wrap my head around the code logic for something like that.

abettermap:
also forgot to reply to your comments:

> For clearing, does the user need to click away or refresh or something? 

No. why, is not working for you?

> I wonder if there should also be a note, like right after "You can clear or view them if needed", saying "You can also set a filter to view a specific set of communities on the map". 

hmm if you're referring to filtering-filtering in the table, wouldn't they already know that that's an option since they've already done it? 

if you're referring to filtering-ish-via-Explore then yes i think a blurb would be good. i'd like to adjust the wording tho as it implies table filtering to me. would also ideally link to `/Explore` there as well.

rperlin-ela:
Ok, thx, let’s leave the Neighborhoods alphabetized — unsorted if primary/secondary can’t be distinguished probably just confusing.

Right now yellow dot is all good and clearing filters is working well!

Blurb-wise, maybe we’re talking about different things. I’m imagining someone who just comes to Explore and doesn’t even know about Data and pulls up all the Western Asia or whatever languages but then wants to see them mapped and sees the map not changing. So I guess this is different from the current blurb only for people who have filters on.

abettermap:
Oh gotcha. How about a 24/7 blurb for that at the top of the main /Explore panel then? Wouldn't need to show it on any of the sub levels, that might get obnoxious, but I think a top-level "Curious about filtering? Talk to your doctor about Data™ today." link would be encouraging. 

As for the neighborhoods I will just sort alphabetically then. I understand from your perspective and your data intimacy level why things like this and the primary/secondary distinction in general are important, but honestly (and speaking as an ignorant user over here, and not even that ￼￼￼ignorant since I've seen just about every record there is), those distinctions feel very trivial, if not overlooked, while I'm using the app. It's very enjoyable and satisfying to not have any dead ends and just being able to click until I'm blind. I would bet money that things which are obvious to you like showing two different neighborhoods for the same community in the last Explore page before Details will go unnoticed by the majority of users. New Yorkers might pick up on it if they're paying attention, or they may be like me just clapping their hands in idiotic excitement. "Clicked on card, things happen, make J happy." 👏🏻

I think more blurbs can really cover some of these bases. They're easy to implement on my end, and can be hidden behind an info button or something. I think something like that could go on the last explore views before details, and also next to the locations list in details. It would just be a little ℹ️￼ button that opens a popout explaining the neighborhood sitch. 

Thoughts?

abettermap:
also found a bug, won't get to fix today but just fyi. anything like this `/Explore/Macrocommunity/Shi'a,%20Sunni`, not playing nicely. probably the comma. probably same problem Town had. on my list to fix, don't change your data.

## unrelated proposal
i may have finally conquered the panel header, close btn, and breadcrumbs:

image

i understand why you wouldn't want breadcrumbs in Home or Details (makes sense to omit in Home for sure), but now that i've got a handle on it, how about that bar 24/7, sans breadcrumbs in Home (and Details if you aren't into it) BUT return the `?` help btn there once again and drop it from the bottom bar? and also move the `i` from the map ctrl btns as well.

reasons:

1. Help feels weird coming from bottom nav. as does Data, but not as bad.
1. `i` btn has nothing to do w/map ctrls and takes up space, kinda busy. i think a small icon to go with the `?` would fit just fine and be more accessible. 
1. frees up space for another proposal as we have census-ness coming in: remove the geolocate btn from the map ctrls, and into a "Layers" tab, where census things will go as well. the stuff in the current geoloc/geocode gets somewhat lost in the mix and is not really a map-navvy thing as much as the zoom in/out/home/reset. the menu in there has kind of just become the dumping ground for mappish things, and that's where i was planning on putting census things. also the menu obscures the map. since it does stuff to map, would be nice to be able to see its impacts, e.g. the current location icon AND would open the door to seeing geocode results without having to close the menu (oh if that wasn't clear, the menu contents would go into its own panel). 

so, TLDR

1. help and info btns out of their current loc's and into panel title bar.
1. geoloc/code out of current loc and into Layers tab at bottom.
1. ...where it would be joined by census ctrls, et al

LMK what you think.

## updates

i just pushed another 15 commits up, UI looking really good i think. shouldn't be any more broken stuff (e.g. the English (Irish) scenario) except for commas in URLs like the Shi'a one.

not going to get the unique instance/examples into this go, and likely not in the SOW. will have to settle for the ellipsis or nothing for now (except in top-level Explore, which was easy to code since there is only one level of checks). it was just an idea anyway, didn't mean to make it an all-or-nothing requirement, and it's just a small card footer anyway, not a title or intro. 

let me know if find any problems, calling it for tonight!

abettermap:
oh also did not get to the long-list thing today but i did make cards fit 3-across columns in some screen sizes (too narrow in others). obviously not on phone size, but since touchscreens and touchpads have the inertia/kinetic scrolling, it doesn't bother me as much.

i also shortened the row height, so there's less scrolling that way as well:

image

not to mention there are only a handful of instances where there are tons and tons of cards. even on on /Explore/Neighborhood i can get all the way through in 6 flicks of the mouse wheel on a laptop.

so, really the only monster is /Explore/Languages. and unlike the others, the nested routes within that particular path can be accessed by _many_ means since /Explore/Not-Languages/Something/LanguageName leads to a simpler list.

lastly, if someone has filters on, they will only see a subset of Languages anyway, not the 600+ list.

that said, i'm reluctant to spend time on creating some kind of in-panel navigational basically just for one scenario-  /Explore/Language in an unfiltered setting.

not to broken-record, but... thoughts?

rperlin-ela:
Great stuff all around. Aside from this small stuff below, I think this is really ready for primetime, merging, whatever.

> Oh gotcha. How about a 24/7 blurb for that at the top of the main /Explore panel then? Wouldn't need to show it on any of the sub levels, that might get obnoxious, but I think a top-level "Curious about filtering? Talk to your doctor about Data™ today." link would be encouraging.

Yes, agreed, let's use similar language to what's in the table. Like "To view results in the map, filter columns in the **Data** table" or something like that?

> As for the neighborhoods I will just sort alphabetically then. I understand from your perspective and your data intimacy level why things like this and the primary/secondary distinction in general are important, but honestly (and speaking as an ignorant user over here, and not even that ￼￼￼ignorant since I've seen just about every record there is), those distinctions feel very trivial, if not overlooked, while I'm using the app. It's very enjoyable and satisfying to not have any dead ends and just being able to click until I'm blind. I would bet money that things which are obvious to you like showing two different neighborhoods for the same community in the last Explore page before Details will go unnoticed by the majority of users. New Yorkers might pick up on it if they're paying attention, or they may be like me just clapping their hands in idiotic excitement. "Clicked on card, things happen, make J happy." 👏🏻

I hear you, though I also feel a responsibility to the 5% of users who will really be using this more heavily (we hope) for real-world purposes. Trying hard not to be picayune, I swear, and to make this actually reflect reality and not just be another shiny app :)

> also found a bug, won't get to fix today but just fyi. anything like this `/Explore/Macrocommunity/Shi'a,%20Sunni`, not playing nicely. probably the comma. probably same problem Town had. on my list to fix, don't change your data.

Great catch and glad you are on the fix.

> how about that bar 24/7, sans breadcrumbs in Home (and Details if you aren't into it) BUT return the `?` help btn there once again and drop it from the bottom bar? and also move the `i` from the map ctrl btns as well.

Yes, I like it, go for it.

Aside: I feel like you may have strong feelings, but sometimes here I felt like it was too much with the icons, e.g. on every language card in /Explore/Language), but also elsewhere. Likewise under Language Family and Neighborhood, though not World Region since the colors introduce some variation which is great, but whenever it's just the same icon 100+ times. Would you consider reducing?

I think we can nix the word "Example/s", but maybe just add ellipsis as elsewhere, or ideally at least where there's more to show?

Hopefully no trouble for you but I feel like Status is the weak link category and I think we should take it out altogether. "Explore" is in any case clearly a curated group of things, not the full table.

Unfortunately I'm getting the yellow-dot problem again on Safari desktop. It's any time, in full screen on my big (dual) monitor, that I set a filter. As I shrink the window it goes away. As I expand it towards full screen, the dot suddenly expands and smooshes. Not a problem in Firefox.

Btw, still having that issue with Help, maybe when the window's been opening for a while, where nothing shows up and then it works when I refresh.

abettermap:
All good points, thanks. I broke out the last comment into a new issue and will merge the existing changes in a sec. 

Re: The 5%: you'll have to break it to your advanced power linguist real world users that some of the shortcomings not apparent to the other 95 are unavoidable in the current state. I demoed the app to one of those 95ers tonight and she found it very intuitive with the exception of the primary/secondary neighborhoods (multiple countries was fine). Maybe she's biased being a GIS gal, but she felt that each of those language-neighborhood instances, however non-unique they may be, should have their own record in the dataset so that they each lead to their own point. I probably didn't convey the reasoning for this very well but that was the strongest sense of confusion I gathered after our 2-3 beers. 

Agreed with the excessive icons on the cards that are repeated, and also about not having examples text prefix. I would like to keep the little language icon in the footer of those cards though, repetitive or otherwise, since there is no text to back it up. The fact that the footer denotes a language might be obvious for some instances and some users, but I don't think that tiny little reinforcing icon is hurting anything. Will definitely remove it from the card titles though￼￼￼, it's a bit much. 

I will also drop the examples prefix in the card footer but unfortunately have to keep deal…. If I knew when there were more instances than I could get the totals and also be instances, so in love being able to figure that out I felt like the ellipsis Was the best option even if a little misleading. ￼￼

No problem removing Status, now I don't have to add icons. :)

As for the big dot, can you provide me with your screen resolution/dimensions somehow? If you're on a Mac it's system preferences then display I believe. I'm guessing it's something between 1300 and 1920 but if not then we might have to screen share￼￼￼. 

Help and About issue, aware of. Haven't had time to address amongst the rest. Always feel free to create issues for bugs, saves me time and makes the problem more real and evident, and helps with my limited supply of focus.

rperlin-ela:
Ok, all good. 

In terms of screen dimensions, it happens on both monitors 13.3-inch (2560 x 1600) and 24-inch (1920 x 1080), but I see now that it happens specifically when it transitions between a desktop-style setup with the panel on the left to a mobile-ish setup where the panels are on the bottom.

Thanks for your friend’s feedback, and I totally get that and it’s something that’s been bothering me for a while. However painful it would be, I think sometimes about trying to figure out a point on the map (unique record) for everything (i.e. no secondary neighborhoods) just to get that consistency, even at the risk of totally overwhelming the smaller languages and making the map a lot busier. Of course we’d go from 1000+ records to a lot more, but otherwise do you see any impact from your end if this happened down the line and it was all 1-to-1?

I guess if we’re axing Status (and Size) under “Explore”, then the legend blurb “Click any of the items in the legend…” should be changed. Just something like “Click any color-coded region to see languages from there spoken locally”?

abettermap:
> I guess if we’re axing Status (and Size) under “Explore”, then the legend blurb “Click any of the items in the legend…” should be changed. Just something like “Click any color-coded region to see languages from there spoken locally”?

i think it should just be shown only for World Region. then i'd say something like "Click any region to see..." and i'm open to wording on that.

> However painful it would be, I think sometimes about trying to figure out a point on the map (unique record) for everything (i.e. no secondary neighborhoods) just to get that consistency, even at the risk of totally overwhelming the smaller languages and making the map a lot busier. Of course we’d go from 1000+ records to a lot more, but otherwise do you see any impact from your end if this happened down the line and it was all 1-to-1?

UI-wise, don't think so. it would even encourage more exploration since each card would lead to a new map location. yes there would be inconsistency since some cards still lead to only secondary neighborhoods, but it's still an upgrade from up to 5-6 of them leading to same hood for one point.

code-wise, i don't _think_ so. i'm sure there could be some small tweaks to make but since it's set up to handle comma'ed or non-comma'ed values, it "should" just work. the comma support needs to stay regardless since Countries has it (and that's working great).

### short-term
what i would do before you go to town (errr neighborhood) on this, is not go all-in with breaking out the primary/secondary for ALL records, but rather just be more liberal on the ones you may have originally considered breaking out into multiple points but ended up as secondary/tertiary, etc. Give those guys their own points instead (whenever you can justify doing so and it's not TOO much of a stretch).

if the concern is certain communities undesirably obscuring smaller ones, have you tried sorting the spreadsheet in a different order prior to export/upload? i'm not aware of MB sorting by ID (don't quote me on that but i can't see a reason why it _would_) so in theory you could control the order of that. 

what you could try is just cutting a spreadsheet row for a record (say Row 22) that you know is showing up as a point beneath another point in the map, then paste the cut row above or below`*` the one showing above it in the map

`*` _i'm not sure if the sheet order is same as MB order or reverse, so just swap the relative above/below-ness of the row relative to the other row_

hard to explain so let's say sheet looks like this:

```
Row 22, whose point is drawing below Row 47's point in the map
Row 47
```

...you would cut Row 47 and paste it above Row 22, then export, upload, and see if it made a difference in the order.

### mid-term
seems too easy but if that DOES work, then what i would do is add a "Symbol Priority" column with values as numbers 1-10 or something (or blanks for ones that don't matter and/or aren't causing obscurity), then you could sort by that column priority to export.

this would throw off the alphabetized list but sorting initially by `ID` or `Language` is a much easier problem to solve on my end than "last resort"

### last resort
1. you do the Symbol Priority column
1. i would use some MB settings to draw them in that order (not in this SOW though, that could take some doing!).

### if obscuring is not a problem
...and i totally misunderstood, then just disregard most of this comment.

abettermap:
also, Re: my "primaritize/de-secondaritize some but not all" suggestion, if you go that route perhaps we should not include secondaries in the cards? might lose some exploration that way and also would have to factor in how that would affect the secondary neighborhoods in Details. would have to change the logic so that it's "if this neighborhood only occurs as secondary, don't make it linkable". no dead-ends that way, but it might be confusing to user with:

1. current hood: bold, linkable
1. add'l hoods: linkable but only if they are for a community for which they are the primary.
1. others: non-linkable
1. Towns: non-linkable

getting purrrrrdy crazy there, but probably not as bad as as it sounds code-wise. in fact, much easier because i'd only have to check the first item in the split/parsed out comma'ed list. although then it would get weird because i'm already doing that for Country, so we'd end up with three scenarios: 

1. commas, but only the first split item matters for the linkability check
1. commas, all items split are linkable
1. no commas

hard to explain/think about and probably not happening any time soon, so i'd say just start breaking a few out into their own records wherever possible. users can still discover other hoods for a given language via Details if they click that language name's chip. it might be even more likely that it gets clicked than the neighborhoods since they aren't chippy, and because Language chip occurs first.

rperlin-ela:
Many thanks— this is all good stuff for Future Ross. I’ve previously done some breaking out of additional neighborhoods into separate records, where it feels justified enough and where we can scrounge a worthy lat-lon, but will try to do another sweep (not this week). There are a few issues, of which obscuring is one, and it’s useful to know there are options with that, and comforting to know that going one-to-one at some point won’t kill the UI or the code

Not having cards for the secondaries seems like a good idea, actually, certainly given the current state of play where they take you somewhere potentially unexpected and don’t do too much else. But I guess best to hold off on axing for the moment what with everything else there is to do.

Issue #132, opened 2020-11-08
Wishlist for Panels/Explore


abettermap:
## Description

* [ ] Show unique item count and examples for all relevant scenarios in each card, like so:
    ![image](https://user-images.githubusercontent.com/4974087/98485581-b6e24e80-21d4-11eb-822c-e3e5bf14b5b1.png)
* [ ] Panel transition/animation on route change (opacity is low-hanging fruit, start there and mv to translateX if it's not insane)
* [ ] "Maximize" panel option on mobile. Probably related in one of the outstanding Wishlist issues, but basically have a third panel state that obscures the map or nearly so in case user wants to see all/more of the panel's contents.
* [ ] Remember vert. scroll position on panel/route change (instead of defaulting to the top. 
* [ ] Bottom nav click for current route:
    1. e.g. /Explore: panel already at top-level view, so close the panel
    1. e.g. /Explore/World Region: click should go to /Explore

### Even more wishlist: long card list improve's
* [ ] react-window for perf
* [ ] some kind of nav to get through the long lists
* [ ] ...coupled with a back-to-top btn

will add more as needed, can't read my own notes.

rperlin-ela:
Sorry if this is the wrong place for an old discussion. And this may be a wrong-headed question on something we've already spent way too much time on 

But in a case like Explore/Neighborhood/BayRidge (or any neighborhood), I think it's wired up to show only that same Primary neighborhood in the card and any Secondary neighborhoods. Understandable why we did that, but I think especially now with the enhanced Lg Profile page that it can list (only) all the Primaries just like the other "pre-Language" pages do. See attached screenshots.

![Screen Shot 2020-12-08 at 3 47 06 PM (2)](https://user-images.githubusercontent.com/52076788/101540673-5890c300-396e-11eb-8294-fb75728c06a2.png)
![Screen Shot 2020-12-08 at 3 52 16 PM (2)](https://user-images.githubusercontent.com/52076788/101540675-5890c300-396e-11eb-9307-ad3c4e133707.png)
![Screen Shot 2020-12-08 at 3 54 02 PM (2)](https://user-images.githubusercontent.com/52076788/101540676-59295980-396e-11eb-9637-3da2f4ef89fa.png)

PR #133, opened 2020-11-10
Explore wrapup, other improvements and bug fixes

### LINKED ISSUE #92: Table: show horiz. scrollbar on desktop sooner
## Problem 

Can't really tell it's there:

image

until you're all the way to the bottom:

image

Mobile not a big deal, it's obvious and easy to scroll. 

Tried the `doubleHorizontalScroll` property but not helping much: https://material-table.com/#/docs/all-props

## Solution

This one's tricky. Maybe one of these?

1. CSS to absolutely/sticky position the bar?
1. If not, something overkill like floating right/left arrows that's enabled/visible when you're not scrolled to each end.

### LINKED ISSUE #98: Panels on map: miss u buddy 😞

### LINKED ISSUE #101: Table: don't load until needed
This includes alllll the stuff that comes along with it, like flags:

image

It might be the `keepMounted` setting in the `<ResultsModal>` but either way it seems unnecessary to load all this stuff prior to when the table is opened the first time.

### LINKED ISSUE #112: Fix meta typos in index.html
image

Search project to make sure no others.

### LINKED ISSUE #128: Explore wrapup
## Source
__Originally posted by @rperlin-ela in https://github.com/Language-Mapping/language-map/pull/127#issuecomment-723375179__

### Bugs
* [x] Support commas in URLs, e.g. `/Explore/Macrocommunity/Shi'a,%20Sunni`. **I am no longer seeing this comma'ed instance, only Shi'a and Sunni separately, so calling it resolved.**
* [x] Correct Vatican City flag (see #130)

### Panels/breadcrumbs
* [x] Panel title bar 24/7, sans breadcrumbs in Home (and maybe Details) 
* [x] ...and return the `?` help btn there once again and drop it from the bottom nav.
* [x] and also move the `i` from the map ctrl btns as well.

### Card lists
* [x] can nix the word "Example/s", but maybe just add ellipsis as elsewhere, or ~~ideally at least where there's more to show~~ **ideally yes, makes the most sense, and i gave it a shot but will have to call wishlist on this.** so `...` 24/7 for now, better than empty space. will keep an eye out for solution to reach the ideal. Added it to wishlist in #132 
* [x] Card lists: rm card header icons where too redundant, e.g. on every language card in /Explore/Language. Likewise under Language Family and Neighborhood, though not World Region or Country since the colors introduce some variation.

### Size/Status
* [x] Remove from Explore
* [x] ...and make them not clickable in legend

### Blurbs
* [x] ...then update legend wording and only show if World Region is the symb type. Might as well include the "Show world map" toggle as well.
* [x] Add blurb in Explore panels to encourage users to view Data:
  > Yes, agreed, let's use similar language to what's in the table. Like "To view results in the map, filter columns in the **Data** table" or something like that?

### LINKED ISSUE #129: "Help" not loading
## Summary

"Help" seems not to load/appear sometimes, possibly when the user has had their screen open for a while — clicking "Help" just returns a kind of semi-blacked out map. On refresh, it loads fine.

### Screenshots

Screen Shot 2020-11-07 at 9 12 22 AM (2)

### LINKED ISSUE #130: Chips with US flag oversized
## Summary

Small issue, only on Safari desktop from what I've seen, but any chip using the US flag on the Details panel seems to be oversize, placing "United" and "States" on different lines. Doesn't seem to be a length issue, since it doesn't happen with United Arab Emirates, for example.

Screen Shot 2020-11-08 at 9 38 31 AM (2)

### Filters indicator dot

I think this may be related to the [squished dot](https://user-images.githubusercontent.com/52076788/98261442-7b385200-1f52-11eb-9b7f-c30025c2a425.png) problem (more in https://github.com/Language-Mapping/language-map/pull/127#issuecomment-722487052) . will try to fix in one go.

### Vatican City flag missing
Totally different problem but might as well lump it in: 

image

i think i just need to change this to Vatican City:

image

### LINKED ISSUE #131: Non-Explore bug fixes for SOW.2
## Summary

* [x] **Fix map offset on mobile:** screwy in all zoom-to-bounds scenarios. and in pitch reset. seems to be off by consistent amt for bounds scenarios and clicked points (although that works in our favor since the popup centers better vertically).
* [x] **Page title should reset** after coming from Details, but it's still at the previous:
    image
* [x] **Gentium ~~Safari error~~ console warning:** At least look into it if it's an issue. Something about ttf MIME type: 
    image

abettermap:

* Closes #129
* Closes #98
* Closes #128 
* Closes #130 
* Closes #131
* Closes #112 
* Closes #92 (you're welcome, @markturin 😉 )
* Closes #101 I guess? Not really getting it how I wanted but that might not be doable since it's not really behind a route due to the table state requiring `keepMounted`, etc. Probably good enough to close, if nothing else to prevent future distraction and to serve as a `React.Suspense/Lazy` reference.

## Other additions

I took some more design liberties here, hopefully that's cool. The big ones being more panel title bar complexities, and colored panel title bar and bottom nav. The latter will eventually house the Layers/spatial stuff, so we'll see how it looks with a fifth btn.


abettermap:
testing as i go along, it's pretty sound. probably some tweaks to make but i believe this PR completes the SOW2 high-priority items. feel free (anyone) to document any problems!

abettermap:
I also forgot to mention I got the unique/examples instances card footers from #132 

i opted to leave icons intact for some levels for users unfamiliar with the data:

image

but made sure to remove them from the super-excessive spots:

image

image

or a mix of both where it made sense to me:

image

PR #140, opened 2020-11-30
Census: switch to Sheets approach

abettermap:
Use google sheets as the census attribute source instead of baking it into the spatial.

### DIFF
--- a/src/components/map/CensusLayer.tsx
+++ b/src/components/map/CensusLayer.tsx
@@ -1,48 +1,71 @@
 import React, { FC, useState, useEffect } from 'react'
 import { Source, Layer } from 'react-map-gl'
-import { FillPaint, MapboxGeoJSONFeature, MapSourceDataEvent } from 'mapbox-gl'
+import { FillPaint } from 'mapbox-gl'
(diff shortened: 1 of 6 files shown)


abettermap:
@rperlin-ela fixed the oil spill and am now using the two 2013-2017 Sheets workbooks directly, so @fiddleHeads might officially be off the hook for census GIS stuff as there are now only two fields in GIS and they never change:

image

So Ross, between those two workbooks and the Languages Config, you should have total control over the census stuff now. Let me know if you run into any problems!

abettermap:
Shoot, I forgot that production is using all the MB fields still, so that's likely breaking right now since I removed them￼. So we should get this merged ASAP as soon as it's been tested a bit￼.

rperlin-ela:
Not sure how much or what kind of testing is needed, but looking at deploy 140 quickly and trying out different census data layers, everything seems to check out

abettermap:
If you're not seeing the black oil spill like we were before when you zoom in close and then turn on the layer, I'd say were good. Merging.￼

rperlin-ela:
Awesome!

PR #145, opened 2020-12-07
Pre-details wrapup

### LINKED ISSUE #144: "Pre-details" wrapup
## Description

Finalize the "language profile" page.

#### Census popout
Remaining wording/UI edits: https://github.com/Language-Mapping/language-map/pull/143#issuecomment-739982298

- Vintage
- Rm btns heading
- Change wording
- Make language bold: _Related ACS 2014-2018 Data (PUMA-level): **Tibetan**_

#### Language-level chips
(some of these need to be added to Config sheet first)

- Country
- Language Family
- World Region
- Macrocommunity (not sure if this is language-level)

 **Did we decide if these (and the non-clickable meta like glotto, iso, global speakers) will go into a popout or show 24/7?**

### Body (Neighborhoods)
Described in https://github.com/Language-Mapping/language-map/issues/136#issuecomment-738491655 and https://github.com/Language-Mapping/language-map/issues/136#issuecomment-739440239.

abettermap:
## Summary
Improvements to census popover, wording/blurbs throughout UI, pre-Details. Use lang-level from Config wherever applicable, fall back to instance-level as needed.

- Fixes #144


abettermap:
@rperlin-ela solid progress today, got most of the pre-Details stuff in there (except "Share") and a bunch of UI improvements/consistentizations for the "intro" sections as well. Once it builds, this is a good one to test (probably the only one until Country is populated):

https://deploy-preview-145--languagemapping.netlify.app/Explore/Language/Arabic%20(Algerian)

I have to say i disagreed with bold-ing the census name instead of its label in the popover. A bold heading is pretty standard, and the value it describes is non-bold:

image

or

image

or

image

or

image

If you want me to juggle the UI and/or wording, no problem, it just seems like a good standard to preserve.

rperlin-ela:
Looking great to me.

About bold-ing, I understand what you mean, and maybe it would feel different with the popover in, but I think this is a bit of a different case. I don’t think it would be meaningful to bold everything before the colon, so if you’re very anti-, we can just axe bold here altogether unless you have another way to rework

abettermap:
you're right, i don't think full pre-colon bold works here so i just went all-in with a different approach, what do you think?

image

the largest monster of all is this one, and it's still not terrible:

image

rperlin-ela:
Approach is good. Just need to clarify and hedge the phrasing a bit because these connections are our based on our analysis, so how about 

“Speakers of Arabic (Algerian) are likely to be included within the census category of Arabic, for which ACS 2014-2018 data is available at the tract level." 

And maybe “View all census language data”?

abettermap:
## Replies

> “Speakers of Arabic (Algerian) are likely to be included within the census category of Arabic, for which ACS 2014-2018 data is available at the tract level."

k how about this?

image

> And maybe “View all census language data”?

hmm, we aren't really showing data in the "data" sense like we are in the table, and either way the link just takes the user to _options_ for interacting with it, so how about:

1. View all census language options
1. View more census language options
1. View all available census languages
1. View more census languages
1. View additional census languages
1. View additional census options
1. Click here to win a prize

**Alternative words**
(can replace "view" with "show" or "display" or "go to" or "see")
(can replace "options" or "languages" with "categories")

## Updates
ok gang's all here i think? if we add anything else, she's gonna burst, captain!

image

i mean it'll "fit" but once any wrap-causing multi-country chips, descriptions, and lengthy language names start crashing this UI party, mobile users might not know about the Neighborhoods gold that lies below. and i'm reluctant to downsize any further, it's got a good snug fit as-is. the only thing i might do is reduce the height of the Descrip preview, although i've already done that a few times. mobile users are used to scrolling, i'm sure they'll fine the treasure below.

is there anything else to add or can we call this branch good (after your replies to Replies)?

abettermap:
(btw this whole "Language profile" thing we've got going is pretty slick! much more than the bare-bones Explore's we started with)

abettermap:
i pushed up what i settled on, should build/deploy shortly. 

also i added Video/Audio cols to Config, and populated Acehnese with video for your testing needs.

LMK what to change, otherwise i think we can wrap up this branch?

rperlin-ela:
Banging — I think we can wrap up!

One thing when you click “Share”. I know it makes sense to reuse text, but I feel like this needs a tweak. Like can we slightly modify “Share this Arabic (Algerian) community” to (I’m not totally sure but) maybe just “Share Arabic (Algerian) in New York”? Or “Share this description of Arabic (Algerian)”?

Agreed that these Explore pages have come a long, long way!

abettermap:
Yeah I can totally set it to that text for Share, it's currently just using the default I set when you share from the sidebar (as opposed to Details). 

However, it's actually set to use Description (language-level), but I know not all languages have one, so yeah I'll have it fall back to your suggestion and use Description whenever available. 

Done with computers for the night, will wrap this up tomorrow.

abettermap:
Oh you're talking about the text above the buttons. I was referring to the summary text when you actually click one of them. But yeah, I can change the above-buttons text as well.

rperlin-ela:
﻿Cool, and language-level for the summary text but falling back to instance-level otherwise would be great

abettermap:
Reminder for Near-future Jason: make sure the census dropdown value persists between view changes.

abettermap:
> Cool, and language-level for the summary text but falling back to instance-level otherwise would be great

So for this, what if there is more than one instance of that language? Will they each have the same instance-level description and it doesn't matter which one I use? ￼Or, if there's more than one instance can I expect a language-level description to be available?

We are just talking about the summary used in the share buttons, but I assume this applies to the pre-Details "read more" description as well?￼

abettermap:
> One thing when you click “Share”. I know it makes sense to reuse text, but I feel like this needs a tweak. Like can we slightly modify “Share this Arabic (Algerian) community” to (I’m not totally sure but) maybe just “Share Arabic (Algerian) in New York”? Or “Share this description of Arabic (Algerian)”?

how's this?

image

rperlin-ela:
> So for this, what if there is more than one instance of that language? Will they each have the same instance-level description and it doesn't matter which one I use? ￼Or, if there's more than one instance can I expect a language-level description to be available?

Ultimately, if there are multiple instances and they differ, there should be a language-level Description. I think we're 95% of the way to that goal, but I think there may be a few that aren't yet (e.g. African-American English) and I can't think of an immediate way to check the rest without going through the whole list. I'll try to nail it down soon. In the meantime, I'd say it doesn't matter which one you use.

> We are just talking about the summary used in the share buttons, but I assume this applies to the pre-Details "read more" description as well?￼

Makes sense, assuming we want to fill in something for this across the board. So if there's no language-level and only one type of instance-level, just repeat what's in Details?

Separately, should have mentioned this before when discussing order, but maybe Global Speakers before Glotto and ISO? Not a big deal either way, just a thought

rperlin-ela:
> > One thing when you click “Share”. I know it makes sense to reuse text, but I feel like this needs a tweak. Like can we slightly modify “Share this Arabic (Algerian) community” to (I’m not totally sure but) maybe just “Share Arabic (Algerian) in New York”? Or “Share this description of Arabic (Algerian)”?
> 
> how's this?
> 
> image

Looks good

abettermap:
> Makes sense, assuming we want to fill in something for this across the board. So if there's no language-level and only one type of instance-level, just repeat what's in Details?

Details as in what's currently the instance-level Description, show below the intro stuff? Yes if i'm understanding correctly. I think we decided though that I wouldn't have to do any checks to see if a lang-level descrip OR instance-level exists AND see if instance-level count is 1?

Either way, i need to be careful with this if we're showing lang-level descrips (when available) in the up-and-coming "Read More" of Details' intro, but at the pre-Details view I got the instance-level fallback working no problem, but i forgot what we decided on the redundancy with the card footers?

image

we're kind of running out of stuff that's unique to Details, it's mostly covered in the pre-Details now, but for sake of UI consistency i'd rather not leave the card footer empty, so how about something simple like:

image

or maybe _Click to show in map_ makes more sense?

> Separately, should have mentioned this before when discussing order, but maybe Global Speakers before Glotto and ISO? Not a big deal either way, just a thought

sure thing, easy fix.

abettermap:
i'm blunt-and-deliberate mode lately, how's this?

image

rperlin-ela:
> > Makes sense, assuming we want to fill in something for this across the board. So if there's no language-level and only one type of instance-level, just repeat what's in Details?
> 
> Details as in what's currently the instance-level Description, show below the intro stuff? Yes if i'm understanding correctly. I think we decided though that I wouldn't have to do any checks to see if a lang-level descrip OR instance-level exists AND see if instance-level count is 1?

Right, sorry I meant instance-level Description. And I don't want to make you run more checks. But if instance-level fallback is working, it seems like we're in good shape.

To review (so that things are straight in my own head), I see the following scenarios with Descriptions, but let me know if I'm missing something:

Distinct Lg-Level and Distinct Instance-Levels: Show each at their level
No Lg-Level and One Type of Instance-Level: Show instance level at both levels
No Lg-Level and Multiple Types of Instance-Level: Show any instance level. I'll try to write a new Lg-Level soon.
Lg-Level but No Instance-Levels: We're not going to allow this for now

> or maybe _Click to show in map_ makes more sense?

Yes, assuming the Primary/Secondary split is on the way, I like this.

> i'm blunt-and-deliberate mode lately, how's this?
> 
> image

How about a middle-ground option like "Click to show in map and display details"?

abettermap:
> Distinct Lg-Level and Distinct Instance-Levels: Show each at their level
> No Lg-Level and One Type of Instance-Level: Show instance level at both levels
> No Lg-Level and Multiple Types of Instance-Level: Show any instance level. I'll try to write a new Lg-Level soon.
> Lg-Level but No Instance-Levels: We're not going to allow this for now

not sure what you mean regarding "Show each at their level" and similar. do you mean in the pre-Details ("language profile") card footers and Read More? maybe just pick a handful of real-world examples from the data/config (assuming it's current enough to do so) and let me know what those languages are and what their pre-Details would look like.

> How about a middle-ground option like "Click to show in map and display details"?

that works.

rperlin-ela:
No worries. I think I'm clear, and if you're clear on things, we should be good.

abettermap:
i'm not, but it's either down to this:

image

or this:

image

the second one is already set up and functioning, so let me know if that's correct.

abettermap:
second one, applied to multi-instance:

image

rperlin-ela:
Second one is correct.

rperlin-ela (approved):

abettermap:
@rperlin-ela ok i think we're at a good point to merge. the Neighborhoods stuff, possibly including your comment: https://github.com/Language-Mapping/language-map/issues/132#issuecomment-741025608

in that branch i would also like to use language-level Config instead of instance-level wherever possible in /Explore . Should be able to do most or all of that except for pre-Details and `/Explore/Neighborhoods`. REALLY wish we had a way to do that, starting to question my "Primary + Secondary combined into one column" suggestion. The parsing thing is not terrible with Country, but throwing on the addl concept of "the first one before the comma is special" is quite a little nightmare in the code, sometimes i can't even walk through my own logic that i've already written! i'm sure code just looks like code, but trust me that this is nasty code:

image


might try to brainstorm some spreadsheet sorcery that would give us a pre-populated "Primary Neighborhood" column in Config, which would be much easier to work with than the primary/secondary parsing shenanigans.

just thinking out loud but would it be possible, theoretically, to have a field called "Location", which is the primary neighborhood or town, and then "Other Locations" instead of our current approach? our current approach made more sense before all the Explore and pre-Details stuff joined the party.

rperlin-ela:
I’m not against re-splitting Primaries and Secondaries if it’s useful and maybe even merging Neighborhoods and Towns, though the latter feels messier — but probably not a big deal especially if it’s only on the Config sheet

abettermap:
cool. i think it would probably be a gradual approach, like keep Neighborhood intact first but add Primary so I have something to work with. should be an easy formula:

```
=LEFT(A1,(FIND(",",A1,1)-1))
```

rperlin-ela:
Go for it any time if you want, I’m confident we can always revert if things go horribly wrong. Or Matt or I can get to it soon

abettermap:
i'm fine editing the Config but i'd rather not mess with your data-data. or are you suggesting i do it ALL in Config? that's an idea but i think i'd want some kind of intermediate importrange sheet first, where Primary gets pulled out and then i have something to use in Config for the pre-populated Neighborhoods column. this concept was very handy with the "Census Pretty" field so i don't have to rely on a second or third sheet, it's already in Config.

rperlin-ela:
Ok there’s a Primary Neighborhood column now, let me know what else you need

abettermap:
awesome, thanks for the quick turnaround. if you're okay with it i'd say let's include that in the next MB upload and I can use it easily in the Details chip.

on the language/Config level, i was able to "roll up" the instances into the Config, with a `|||` parser so we don't have to be afraid of commas anymore. do these look right at a glance?

image

interesting formula to make that happen ([credit](https://stackoverflow.com/a/39399630/1048518)):

image

i think i need some kind of `IF` clause in there to avoid things like this:

image

and this:

image

but once that's set, i should have absolutely no conditions to check for in the code (including value existence since many already lack Neighborhood), just a parse on `|||` at that point.

btw one addl benefit i could see with the extracted Primary column is more **validation**, as i think Neighborhood is one of the only cols that can't be validated. i'm sure there's a way but it seems pretty nasty to validate the current Neighborhood col with all its commas, but a one-per-cell would be simple. so, i added another LUT to the Config, `LUT_Neighborhood`, in case you/Matt want to do something with that.

i'm sure there's a way to validate the eventual `Addl Locations` column (ha, i think we've come full circle on this approach), but primary is the most important.

## Bearings/recap

Just a recap on why the Neighborhood-in-Config is useful: now i can use Config instead of instances for all 6/6 of the options at the top-level and second-level `/Explore` levels (except for /Explore/Language). 

Interesting that Neighborhoods = 180 in the current UI, but only 161 in the unique instances of them in the LUT i made. i assume this is because the former includes non-Primary?

image

rperlin-ela:
Nice, at a glance these are looking right. 

All good stuff to know too, thanks. 

It makes sense that there would be some neighborhoods which appear only as Secondaries.

Just uploaded new tileset to MB with Primary column

PR #154, opened 2021-01-06
Migrate remaining Sheets to Airtable

abettermap:
1. MB fonts
2. Census data
3. Census config

**Once done, be sure to clean up Airtable.** Once merged, will no longer need a couple of the tables (tracts, puma) and a bunch of the fields in Language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,31 +49,12 @@ place search, follow these steps:
 **TODO:** document `REACT_APP_MB_SECRET` or whatever for secret scope (aka
 fetching stuff, not MB GL access).
 
-### Google Sheets API
+### Airtable
 
-This project uses Google Sheets to store the config for the dozens of languages
-whose endonyms rely on custom fonts (uploaded manually to the Mapbox account
-used above) in order to render properly as labels in the map.
+~~This project uses Google Sheets~~ **Airtable** to store much of the config and
+some data for the project.
 
-In order for the config to be accessible:
-
-1. Obtain an API key. The detailed steps will not be outlined here since they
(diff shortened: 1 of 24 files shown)


abettermap:
@rperlin-ela if you have a minute, once this deploys give the census stuff a test. nothing new, but i've done the first steps of migrating to Airtable (waaaay easier to work with than Sheets). it's using the "Census" table i created in AT. once we get all this prepped and merged, i can delete the puma and tract lookup tables and all the fields pointing to them.

next step will be to use the actual census data from Airtable. the current deploy is still hitting Sheets.

rperlin-ela:
Sorry, just got to this, but looking fine

abettermap:
ok the upcoming deploy in a few will be **Sheets-free**. i'm hitting the font stuff (for endo labels in map) and census data from Airtable now, so you'll want to test those.

the tricky part was getting it to load without delay since AT has a 100 records per second limit, so for the tracts data it was like 4 seconds for the first load by default. however, i didn't want to add _more_ to the map stuff on load since there's A LOT going on, so my loose hack was to load it just-in-case user does census stuff in:

1. explore
2. details
3. or local (obviously)

it's basically loading behind the scenes, but it would be nice if didn't have to load at all until user interacts w/the census stuff.

anyhoo, i'm ready to merge once you've greenlighted, then we can say adios to Google Sheets.

rperlin-ela:
Poking around and everything so far is looking all good to me, so green light from here. Sheets free!

abettermap:
Woo. Merging. 

I'll see if I can tackle the two outstanding bugs tonight.

PR #155, opened 2021-01-07
Small bug fixes

### LINKED ISSUE #152: Support single-quoted Airtable filter values
## Summary
these guys are causing trouble w/AT API filters, which need single quotes wrapped around values:

image

Might be [this bug](https://community.airtable.com/t/using-api-to-filter-records-fails-when-the-value-contains-a-comma/20035/4) (???)

### LINKED ISSUE #153: Set census from table details view
## Summary

Setting census from the in-table details view does not work. 

## Expected behavior

We haven't really discussed what happens here, but it seems like taking the user to the "Spatial" panel would make sense. This is not necessary in other views because the map is already visible.

## How to reproduce the problem

1. open a full-screen detail view from a record in the table
2. click Census chip
3. click "Show in map" btn

## Related items

#151 touches on it

## Resolution

If not easy, just hide Census btn in that view until we find a solution.

abettermap:

- Fixes #153
- Fixes #152

@rperlin-ela let me know if either of the above haven't been fixed in the deploy.

## Other

Also guarantees that the "Try one at random" btn will find a match (if you take your sweet time waiting for the table to load behind the scenes). No need to test, i've kind of accepted that it may or may not find a match if the user doesn't wait the ~4 seconds and happens to reach this view before then.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -13,6 +13,7 @@ import { paths as routes } from 'components/config/routes'
 import { AIRTABLE_API_KEY } from 'components/config'
 import { AppWrap } from './AppWrap'
 import { wpQueryIDs } from './about/config'
+import { MapToolsProvider } from './context/MapToolsContext'
 
 // // Provide the default query function to your app with defaultConfig
 // const mainQueryCache = new QueryCache({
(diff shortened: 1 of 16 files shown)


rperlin-ela:
All looks good, merge away!

Issue #118, opened 2020-10-17
Table export: enhancements


abettermap:
## Description

Some minor bells to add to the existing whistles:

- Disclaimer/blurb about the missing Endo values being due to font limitations. Maybe mention if user needs to be able to see them then CSV is better option. 
- Indicate that it is only a subset of data with “Showing X of Y total communities” or something 

## Related items

#117 started this ball rolling 

## Resolution

Pain in the butt to add text to the PDF, or at least to position and style it well, so will need to decide how to use the available real estate.

rperlin-ela:
A few more:
- Prevent "Global Speaker Total", if empty, from displaying in PDF as 0
- Prevent comma or other random mark from appearing where there's a missing endonym, e.g. Chinese (Mandarin), Coptic, Bantawa etc
- "Location" instead of "Primary Location" to match UI Table?
- Space after the comma where there are multiple countries
- Switch "full source dataset" link from Sheets to https://airtable.com/shrqQo5FJHvhKtffs

abettermap:
i'll give a skim in the code today but should at least be able to update the full src link

abettermap:
moved the new stuff to #158

PR #159, opened 2021-01-07
Table fixes/improvements

### LINKED ISSUE #156: Clickable "Size" and "Status" in UI Table
## Description

This is a tiny thing, but given the great click-to-filter capability with all the columns (obviously except Global Speakers) I would expect the same with Size and Status. Apologies if there's some reason we decided against this long ago that I've forgotten. If in agreement, only fix if this is a very simple and quick thing! Otherwise totally fine as is, feel free to close issue and ignore.

### LINKED ISSUE #158: Table export: additional requests
_Originally posted by @rperlin-ela in https://github.com/Language-Mapping/language-map/issues/118#issuecomment-756193122_

A few more:

- [x] - Prevent "Global Speaker Total", if empty, from displaying in PDF as 0
- [x] - Prevent comma or other random mark from appearing where there's a missing endonym, e.g. Chinese (Mandarin), Coptic, Bantawa etc
- [x] - "Location" instead of "Primary Location" to match UI Table?
- [x] - Space after the comma where there are multiple countries
- [x] - Switch "full source dataset" link from Sheets to https://airtable.com/shrqQo5FJHvhKtffs

abettermap:

- Fixes #158  
- Fixes #156

@rperlin-ela Also hopefully fixes-ish the table Country click issue once and for all. It wasn't fully resolved: if you click on a Country cell but not _directly_ on it, like a little above or below the value, it breaks. My "solution" ultimately was to just jump ship if there's nothing there instead of the whole app crashing. For fat-fingerers it will just feel like nothing happened so they'll click again, hopefully _on_ the value or flag this time. 

I'm sure there's some kind of styling fix but not worth it given the headaches from this column already, and it's an edge case (literally 😆 ) anyway.

The Country cells have been a total nightmare because they not only contain more than just text, but their source is an array while the table filter column expects text. So many levels of 🤬 . Try in prod to see what i mean then try in deploy to confirm it's fixed, please. Along w/all the exporting stuff.

### DIFF
--- a/src/components/results/ResultsTable.tsx
+++ b/src/components/results/ResultsTable.tsx
@@ -47,16 +47,6 @@ export const ResultsTable: FC<Types.ResultsTableProps> = (props) => {
     const colIndex = tdElem.cellIndex
     const { field } = columns[colIndex]
 
-    // Support multi-line value clicks, e.g. Country, Additional Neighborhoods
-    const elemWithFilter = path.find((elem) => {
-      const asElement = elem as HTMLElement
-
-      return asElement.classList?.contains(FILTER_CLASS)
-    }) as HTMLElement
-
-    // Safari will preserve newlines w/o `trim`
-    const newFilterVal = elemWithFilter?.innerText.trim() || rowData[field]
-
     if (field === 'id') {
       history.push(`${routes.details}/${rowData.id}`) // show feature in map
 
@@ -71,6 +61,22 @@ export const ResultsTable: FC<Types.ResultsTableProps> = (props) => {
 
       return
     }
+
+    // Support multi-line value clicks, e.g. Country, Additional Neighborhoods
+    const elemWithFilter = path.find((elem) => {
+      const asElement = elem as HTMLElement
+
+      return asElement.classList?.contains(FILTER_CLASS)
+    }) as HTMLElement
+
+    // Clicking a country flag cell above or below the content results in no
+    // element match, but we can't use the row value because it's an array and
+    // the table is expecting a string, and we can't convert it to string
+    // because then there won't be a match if there is a newline. 😠
+    if (!elemWithFilter && Array.isArray(rowData[field])) return
+
+    // Safari will preserve newlines w/o `trim`
+    const newFilterVal = elemWithFilter?.innerText.trim() || rowData[field]
 
     const newlyFiltered = columns.map(
       (column: Types.ColumnWithTableData, i) => {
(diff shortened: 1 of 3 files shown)


rperlin-ela (approved):
Looks good. We lost all non-Roman characters in the Endonym column— Greek, Cyrillic, and some others that used to come through— but I guess it's better to be consistent in that respect for now

abettermap:
internet to the rescue. was able to squeeze a few more endo's out. poor Northern Africa, still empty as ever!

let me know if it works and we can merge

abettermap:
also, do you think instead of empty values there should at least be "uh ohhhh, does not compute, friend-o!" or N/A or something? it's kinda misleading leaving it blank, especially when it's in the table

rperlin-ela:
Good endonym-squeezing! 

Some kind of notice would definitely be good— whatever is most doable or makes sense from your end. Ideally some kind of bold sentence at the top like *For non-Roman characters support, export as CSV or use full dataset and then an asterisk instead of the empty values? Without a sentence at the top I’m less sure how to communicate this, but I’m open

abettermap:
oh i mean in the cells themselves, not the page title. that makes sense too, enough that i already put it in the issue 😆 : https://github.com/Language-Mapping/language-map/issues/118#issue-723812769

emphasis on this part:

image

didn't mean to open a can of worms there. if you want something besides blank cells, i'm willing, but adding text to the page takes time and is difficult to do properly.

rperlin-ela:
I understand, it’s a good idea, just saying I’m not sure it’s going to be clear without such a sentence. If you think it’s an improvement and it’s quick, let's try an asterisk or a hyphen (in the cells) instead of a blank, otherwise just as is.

abettermap:
> *For non-Roman characters support, export as CSV

i'm actually prepping the data the same way for CSV, but users can hit Airtable for the full.

speaking of, what if we left the missing Endo cells empty and just changed the title? (pretend it's centered):

image

rperlin-ela:
Nice— let’s do that, but probably don’t need endonym in there, just “includes non-Roman characters”

abettermap:
trying to avoid more netlify builds today so just let me know if this is good before i push and deploy again:

image

rperlin-ela:
bingo, good to go

abettermap:
cool, give it a try and then i think we're done w/table yeah?

rperlin-ela:
Done!

PR #160, opened 2021-01-08
Polishing

### LINKED ISSUE #99: Panels: transition on route change
## Summary
~~Would be cool to have these transitions:~~ **Just fade for starters**

- Going from "Explore" to "Details": Explore should slide out to the left, Details in from the right
- Vice versa for the opposite scenario

## Implementation
`react-transition-group` with MUI and `react-router` seems a good candidate, or maybe even just MUI `<Slide>`. Former: [tutorial here](https://coursework.vschool.io/react-transitions-with-react-transition-group/), but latter has been very smooth for up/down modal slides.

**ORRRR** Framer Motion and React Router. Some excellent YouTube tuts out there, and it's the smoothest animation in all the land.

abettermap:

- Fixes #99


abettermap:
@rperlin-ela let me know what you think of the panel transitions. I think it brings the "modern app" points level up a notch.

Also fixed that annoying thing when something is clicked that causes the view to change but the "hover" kinda sticks around on mobile on the element underneath wherever the click happened. Hard to explain but maybe it'll be apparent on mobile.

It still happens for non-view stuff, like clicking a chip which doesn't result in a new panel view change, but cards, for example, should not have that problem anymore.

abettermap:
and i'll see if i can knock out the boundaries popup issue first thing tomorrow, then the labels-against-census issue if time.

rperlin-ela (approved):
smoooooth

PR #163, opened 2021-01-08
Last min fixes

### LINKED ISSUE #161: Last-minute outstanding issues
## From Ross's emails

- [x] Description in the UI table is not working for me in prod or deploy, just does a little empty gray rectangle. Safari and Firefox desktop and iPad, but checking quickly mobile seems ok
- [x] Sorry I didn’t catch this earlier, but found an issue with the Additional languages while looking at "Explore/Neighborhood/Elmhurst”. I’m looking at ones like ASL, Greek, Greek (Pontic) and I’m like “these aren’t in Elmhurst”, and I don’t see them in AT, so I can only surmise that they’re getting dumped here because they are languages (primaries actually) of East Elmhurst. Same issue with Harlem (bc there’s an East Harlem), Flatbush (ditto), Ozone Park (since there’s a South Ozone Park), Kew Gardens (because there’s a Kew Gardens Hills). Not too many, I think, but you get the idea.
- [x] why are these bulleted Addl Lgs dead ends again? Couldn’t they be linked to the relevant Lg Profile (aka Pre-Details) page?

## Other

- [x] Table export not working in iOS Safari or Edge app. Chrome is fine.
- [x] Apostrophe issue #152 not resolved for things like `/Explore/Macrocommunity/Shi'a`
- [x] And wanted to double-check again about having multiple Macros for a single language. With Wakhi, it works perfectly except ideally Wakhi's Lg. Profile should have two chips rather than one hybrid chip.
- [x] But it's not just Shi'a— I'm still having trouble with Pa'O, Fe'fe etc

abettermap:

- Fixes #161


abettermap:
@rperlin-ela no idea what happened with the ios Safari thing, we haven't changed export stuff, and i even tried creating an empty pdf, but it still didn't work.

the only workaround i could find was to check if it's iOS, then open the pdf in the current tab. not ideal, but at least it works and "Back" should too. desktop should be same as before.

i think i fixed all the stuff in #163 but let me know if i missed anything in the deploy.

rperlin-ela (approved):
Only two notes in case they resonate:
When I hit Description in the UI Table, it takes a sec and I still see the gray rectangle for a flash before it opens up. 
At Explore/Neighborhood/NabeName before the bulleted languages, it would be good to say "Additional languages spoken in this **neighborhood:** (instead of community)

It's great, looking all good to go. The Computer God at work!

abettermap:
> I still see the gray rectangle 

that would be the modal dialog 😉  Took some detangling but i improved it.

> it would be good to say "Additional languages spoken in this neighborhood"

oh i was thinking it would be generic if it was used in Town, but in that case "location" would be better and maybe that's not a scenario anyway. or definitely not in this round.

changed to neighborhood, deploy shortly. if you see it in the next 15-30, give a test, otherwise i'll merge. would like to tie off with a cleanup branch, including removing the hundreds of country flag files taking up space in the repo.

abettermap:
working for me, merging.

rperlin-ela:
Sorry, was out of commission, but looking now and everything is great!

PR #164, opened 2021-01-09
Cleanup

abettermap:
Big monster cleanup. No impact on functionality or UI, all Typescript and config cleanup.


abettermap:
tested briefly just in case, but changes shouldn't have any impacts

PR #166, opened 2021-01-09
Credits restore

abettermap:
- Restore Country flags credit in README
- Restore 24/7 MB attribution on mobile
- Boundaries popups: stay closed on close


abettermap:
not big enough to test, heading straight to prod

PR #172, opened 2021-01-20
Bug fixes

### LINKED ISSUE #169: Show by "None (hide layer)" crashes the map?
## Summary

Seems straightforward: selecting "None" leads to a delay, then a message that it's loading legend info, and then comes the familiar error (see attached). Tried Firefox and Safari, desktop and mobile, same result. Could it be that we need something in AT > Schema for when the legend is not World Region, Size, or Status?

Screen Shot 2021-01-13 at 7 08 51 PM

### LINKED ISSUE #170: Description missing in Description modal in UI table
## Summary

For all scenarios, Description text, which always worked before, no longer appearing in this modal. Everything else in modal looks fine.

### LINKED ISSUE #171: Making "Explore" categories pick up more exact matches
> * [x]  Sorry I didn’t catch this earlier, but found an issue with the Additional languages while looking at "Explore/Neighborhood/Elmhurst”. I’m looking at ones like ASL, Greek, Greek (Pontic) and I’m like “these aren’t in Elmhurst”, and I don’t see them in AT, so I can only surmise that they’re getting dumped here because they are languages (primaries actually) of East Elmhurst. Same issue with Harlem (bc there’s an East Harlem), Flatbush (ditto), Ozone Park (since there’s a South Ozone Park), Kew Gardens (because there’s a Kew Gardens Hills). Not too many, I think, but you get the idea.

This fix worked, and I wonder if it can be easily implemented across the board. Explore/Country/Dominica picks up languages of Explore/Country/Dominican%20Republic. There are probably very few like this, but it's hard to be totally sure, so just wondering if implementing this logic happens to be a straightforward fix Explore-wide

_Originally posted by @rperlin-ela in https://github.com/Language-Mapping/language-map/issues/161#issuecomment-760471248_

abettermap:

- Fixes #169
- Fixes #170
- Fixes #171

### DIFF
--- a/src/components/explore/MidLevelExplore.tsx
+++ b/src/components/explore/MidLevelExplore.tsx
@@ -71,7 +71,7 @@ export const MidLevelExplore: FC<MidLevelExploreProps> = (props) => {
   const { tableName = field, sortByField = 'name' } = props
   const { url } = useRouteMatch()
   const filterByFormula = prepFormula(field, value)
-  const fields = prepFields(tableName, field === 'Neighborhood')
+  const fields = prepFields(tableName, field)
 
   const { data, error, isLoading } = useAirtable<TonsWithAddl>(tableName, {
     fields,
@@ -113,6 +113,10 @@ export const MidLevelExplore: FC<MidLevelExploreProps> = (props) => {
 
   if (value && field === 'Neighborhood')
     primaryData = data.filter((row) => row['Primary Locations'].includes(value))
+  // Gross extra steps for Airtable FIND issue, which returns in ARRAYJOIN
+  // things like "Dominican Republic" in a "Dominica" query:
+  else if (value && Array.isArray(data[0][field]))
+    primaryData = data.filter((row) => row[field]?.includes(value))
   else primaryData = data
 
   // TODO: better logic for instances, e.g. allow definition
--- a/src/components/explore/utils.tsx
+++ b/src/components/explore/utils.tsx
@@ -60,14 +60,20 @@ export const prepFormula = (
 // Mid-level fields are consistent except a few tables need an extra field
 export const prepFields = (
   tableName: RouteableTableNames,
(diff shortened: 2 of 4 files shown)


abettermap:
@rperlin-ela this should resolve the 3 bugs you posted, let me know if not.

The "fix" for #171 is more like another layer of untreated plywood in my building materials analogy. 🌬️  Yeeikes!

rperlin-ela (approved):
Working on my end, thanks for tackling these!

Before you could do **Show by** "None (Hide Layer)" and still use the Label By dropdown, right? Unless I'm misremembering, I think that was the thing I kind of liked, as a way to see language labels without the dots. Now I'm less sure what "None (Hide Layer)" is good for, just a base map? And even grayed out, it reads "Label by (Language)" so could be a little confusing. If it's not easy to keep the "Label by" options on, maybe it's better just to axe **Show by** "None (Hide Layer)" altogether?

abettermap:
> Before you could do Show by "None (Hide Layer)" and still use the Label By dropdown, right? 

maybe once upon a time when we had a checkbox to show/hide layer and show/hide labels, but we settled on disabling the labels when symb is off. confirmed in code just to make sure it wasn't part of AT shuffle:

image

> I think that was the thing I kind of liked, as a way to see language labels without the dots. Now I'm less sure what "None (Hide Layer)" is good for, just a base map? And even grayed out, it reads "Label by (Language)" so could be a little confusing. If it's not easy to keep the "Label by" options on, maybe it's better just to axe Show by "None (Hide Layer)" altogether?

agreed overall, from a user standpoint anyway, although in GIS it's not directly possible to see labels if their layer is not visible. MB is different concept and possible i imagine, but we kept it out for a reason and it's not related to the newer changes so i'm not going to touch it at this juncture. feel free to make an issue though, it's very helpful for the long-term and you've already got a good GH rhythm going. ;)

merging, glad the fixes worked.

PR #173, opened 2021-01-22
Add Google Analytics via Google Tag Manager

### LINKED ISSUE #108: Set up Google Analytics
## Related
Breaking this out from #90 and #78.

## Reason
So we know how many users are visiting the site, the main benefit being a baseline total against which we can establish a percentage of users experience errors (via Sentry) out of total user visits (via GA).

## Steps
Follow [this guide](https://blog.hootsuite.com/how-to-set-up-google-analytics/) **with these steps/caveats**:

1. You don't need to paste the installation code snippet anywhere. From that step I really just need the Google Tag Manager ID which will look something like `GTM-000000`.
1. Also I'll need the tracking ID from Step 2 that looks like `UA-000000-1` (and so will you for Step 3 in the guide).
1. **Don't worry about anything after Step 3** at this point.

## Future Jason
https://github.com/alinemorelli/react-gtm

abettermap:

- Fixes #108

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,6 +1,20 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <!-- Google Tag Manager -->
+    <script>
+      ;(function (w, d, s, l, i) {
+        w[l] = w[l] || []
+        w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' })
+        var f = d.getElementsByTagName(s)[0],
+          j = d.createElement(s),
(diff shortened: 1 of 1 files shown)


abettermap:
@rperlin-ela I set up GTM and confirmed via their Preview tool that the triggers are firing on page change. Won't be able to see what's in GA unless you add me to that as well. Up to you but these should be the steps:

1. https://analytics.google.com/
2. Click "Admin" at bottom-left
3. Click "Account User Management"
4. Looks like similar steps as giving me access to GTM. Let me know if you get stuck.

Merging, code stuff should be all set.

PR #236, opened 2021-04-02
Counties and census enhancements

### LINKED ISSUE #202: Census: create nested routed panels
## Jason edits

this was originally about census legend and some content stuff but evolved into routing and structure, so i want to preserve the thread below. basically just follow the stuff we settled on in there.

broke this out into #203 and others. anything w/census in the issue title is related.

definitely need to complete #206 first. And don't forget about auto-zoom to selected polygon!

### LINKED ISSUE #210: Bottom nav buttons: link to top-level route for nested views
Scenarios when user clicks Explore bottom nav btn:

1. If they are at /Explore/Anything or deeper, it should take them to /Explore.
2. If they are already at Explore, it should minimize the panel (e.g. like Google Maps app).
3. If they clicked it from elsewhere, it should take them to the last thing they saw in /Explore/Whatever (like it does now), or to /Explore if it's their first time Exploring (again, same as now).

### LINKED ISSUE #234: Bring Counties back
## Data

- [x] Need an Airtable table w/a `name` field at a minimum. **i think everything will come from Maya's CSV derived from GIS and will just be a wholesale upload to create this Airtable table** and will be a one-off. _Maya then Ross_
- [x] Tables using it should point to it (will look into) _Likely Jason_
- [x] ...including the Neighborhood table, which is currently using a Text field for County. yessss more fields, but noooo chances of typos that way. small tradeoff for a new /Explore route ;) _Likely Jason_
- [x] x_min, etc... columns (just making sure to note it, but should be in CSV already) _Maya then Ross_
- [x] MB: either overwrite the existing counties, or start with a fresh Dataset (the latter is preferred but might require 2 uploads since it's likely bigger than the 5MB max upload size) _Jason or Maya_
- [x] an entry in the Airtable `Schema` table (_Ross_)

## UI
(all Jason unless noted)

- [ ] Layer toggle in /Explore/County, just like we have for 'hoods
- [ ] need a County icon _Ross, the icon master_ `CgTrees? RiGovernmentLine?`
- [ ] establish where/how we mention that boroughs are included _Ross_
- [ ] please no chip. too many chips.
- [ ] new color for counties polygons
- [ ] wire up map clicks to trigger route _Jason_
- [ ] column in table (hidden, only added for Searchability), and rm Description 🤩 
- [ ] what else? _Ross_

abettermap:
## Summary

counties are back in business, with their own Explores now, and we are officially MB Boundaries free. census popups and routing-ish are working well with only minor quirks.

- Fixes #234 
- Fixes #202
- Fixes #210 (close enough anyway)

## Remaining

- there's one #210 scenario i'd like to fix but it's pretty minor. i'll try not to bring it up before anyone else does 🤫 
- counties symb needs tweaks
- counties and census selected feature both need to be highlighted
- layer toggles into new menu below the 3D btn
- "zoom to my loc" joins them, removed from Search Locations


abettermap:
@rperlin-ela 

# Many questions
much WIP. but good progress and enough for basis of Zoom so let me know if you're around tomorrow.

big things are that counties don't feel useful as a dediated Explore, the queries for it are not the same concept as Neighbs or Macros or any of the others, popups aren't wired up for counties yet (i don't think there will be a good subheading for those), and i got a good start on making census routeable (e.g. `/Census/puma/F1600__Greek/3604103`), but i'm thinking nested census routes are pretty overkill, at least for changing the panel view nesting. my idea for changing the dynamic symbology based on the URL is not good because we don't want the colors to change when user goes to different URL. so probably better to just show the number of speakers in a popup like we planned. let's forego stuff like "/Census/puma/F1600__Greek" and "/Census/puma" and save it for another scope. lots of work and haven't even touched the legend gradient yet.

rperlin-ela:
Sure, thanks for the update, moving around a bit today but could either Zoom soon (say 9:45 my time) or else probably after 2 pm or so.

Census is definitely the priority so let’s focus on that.  It’s all about just showing the information in a clear and meaningful way, whether that means pop-ups or nesting, with legend. 

We can forego Explore/Counties if it doesn’t make sense or takes us off track.

Seems like all Explore/Something/Something cards, except /Explore/Language are now showing up as endonym-only?

abettermap:
Ross-

solid progress, deploy on its way.

one annoying thing i'm not going to fix any time soon is when you change the census dropdown, it currently has to clear the route in order to clear any popups, which triggers our normally-useful panel animation/transition effect. but it's just annoying since it's not needed here, but i did try to fix it, just saw a rabbit hole happening.

big stuff left:

- ideally highlight the selected census poly. shouldn't be terrible but might save it for later, would rather do higher priority UI things like...
- ...census legend or moving the layer toggles into new menu.
- county field hidden in table
- tooltip on point hover
- #210 (can't WAIT to do this one)

abettermap:
whoops, removed something from timeline code then brought it back, but not quite...

image

rperlin-ela:
Census pop-ups are game-changingly great. Things shaping up nicely!

abettermap:
@rperlin-ela i'm fine to merge this when you are. plenty of remaining small todo's as always, but it was a productive day and it's pretty stable.

abettermap:
ah shoot except for the timeline 🙃

abettermap:
fixed, deploying now

rperlin-ela (approved):
Not testing every tidbit for now, since it sure all looks pretty snazzy to me!

One quick question— is it possible I see the non-pretty census name in the pop-up, e.g. AKDKE003, when I come from the Census pop-up (as opposed to panel)? (Tho I don't mind it for those rare birds who would come by URL)

abettermap:
Affirmative, it's relying on something to be selected in the census search. I was reloading the page a lot during development and it was annoying to see the pretty name undefined, so I figured the non-pretty name￼ was better than nothing. In order to get the pretty name without selecting something in the drop-down￼, It would require a second Airtable query, which I figured was a bit overkill for this edge case. If we had share buttons for that route, absolutely, but I figure most people won't be starting the page from a census pop up. Worst case, they do, it doesn't break and they still see something that at least has the language name kinda sort of in it for puma.￼￼￼

rperlin-ela:
Understood, makes sense!

PR #238, opened 2021-04-03
Gelocation: out of Search into new menu w/other toggles (closes #237)

### LINKED ISSUE #237: Layer toggles: improve availability
this is sprinkled in places like #236 and #234 but the TLDR is to do something like this:

image

to trigger:

image

abettermap:
- Fixes #237


abettermap:
@rperlin-ela i'm good with what i pushed if you are. users are still on a need-to-know basis for neighbs and counties. they can find the source info in Help i guess. the privacy statement for geolocation might be a little more important though, and my downsized "info" btn for it looks better.

approve if you're ready to roll w/it!

rperlin-ela (approved):

abettermap:
also forgot to ask but is this correct?

image

i had the wrong check in there for routes.

also i'll fix the bad popup heading issue where it shows the field name. i thought it was happening only in edge cases but it's always doing it on first click.

rperlin-ela:
You mean the correct text? Yes, I added to it today now that we’ve got all this new Census panel room, but I can cut back if needed when the legend comes in. I thought I might be able to make paragraph breaks or something similar, but I guess no dice?

abettermap:
I can add the paragraph support, I just have it turned off by default￼ because it adds some vertical margins that I have to fight sometimes. Easy to enable though.

As an alternative, and nothing to do with the lack of paragraphs, but would you consider breaking it out into two sections? The less important of the two could go below the controls. 

It's easy enough to implement, so if you're on board then just break it out and create a new record in UI Text. ￼￼i'm just thinking it's already a big enough block of text that I only barely see the search box below it on mobile, and with proper paragraphs and spacing I won't see it at all without scrolling.￼￼

rperlin-ela:
Ok, chopped in two in UI Text, let’s see how it goes

abettermap:
nice. improvement for sure. census just has a lot going on i guess, haha

image

the order looks backwards though. do i have it wrong or you? i put the "top" in the intro.

either way, i doubt most non-power-linguist folks will read even the first paragraph, so maybe we could at least put the toggles above the second paragraph?

abettermap:
for future "UI Text" stuff, be careful about changing id's like that. it's always safer to duplicate a row rather than rename an existing one. renaming the old one made the intro disappear in this panel for example, so just keep it messy AT-wise and we can always clean up later.

PR #240, opened 2021-04-04
Restore hover behavior, improve Details/Profile intros

abettermap:
See commit messages and comments below for details...

### DIFF
--- a/src/components/details/DetailedIntro.tsx
+++ b/src/components/details/DetailedIntro.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react'
+import React, { FC, useState } from 'react'
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles'
 import { Route, Switch } from 'react-router-dom'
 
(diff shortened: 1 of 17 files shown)


abettermap:
@rperlin-ela good progress on these things:

- hover: ELA points were easy, polygon highlights are still a nightmare. if any problems, let's simplify a lot. hard to explain but they selected vs. hover is the main problem. also _cannot_ get it to stop "hovering" on mobile, even though "hover" is prevented for points on mobile. so weird but gotta call it. tooltips are fine though.
- pretty much just look at the commit messages: https://github.com/Language-Mapping/language-map/pull/240/commits

mas mañana...

rperlin-ela:
ELA points hover is heaven, makes me much more likely go exploring around the map. Understood about polygons tho. Also love the new Lg Profile/Details changes. Census text chunks indeed reversed, I can swap them in AT. Toggles up is fine, but final call on all of that should wait till legend.

Onward!

abettermap:
Cool, are we ok to merge then? i have some other census things to bring up, but enough to justify a new issue.

rperlin-ela (approved):

PR #245, opened 2021-04-06
Content bug fixes, new logo

### LINKED ISSUE #86: Help Dialog content improvements
# Summary

Just a few suggestions for the Help dialog, or really any of them that are applicable and WP-driven.

## Screenshots

Instead of full page, only capture the relevant context for that section, e.g.  for Search:

image

Some handy Mac shortcuts for capturing:

1. ...to clipboard: `CMD + Ctrl + Shift +4`
1. ...to new file, likely on your Desktop (new file is probably what you want for uploading to WP): `CMD + Shift +4`

## Table of Contents links

...no longer work. They are correct in the ToC but their corresponding elements no longer have an `id`, so that will need to be applied (hard to explain w/o WP on my end, but Maya knows how):

image

## MB logo

It has a bullet point before it because it's in a list item. Remove the list and just use the image to fix it:

image

### LINKED ISSUE #119: Apply new logo
## Description

It’s still the React logo in some cases like iPad. Fix it!

## Related items

#78 

## Resolution

1. Make a better one
1. Delete the old one/s

### LINKED ISSUE #219: Welcome Screen keeps reappearing in Info
Not a major thing, but for those brave souls who choose "Show on Startup" for the Welcome Screen, everything seems to work fine except the Info panel, where clicking on any level or anchor seems to prompt the Welcome Screen to re-appear, as if we've left the site (becuase we're actually going to Wordpress?).

### LINKED ISSUE #223: Anchors only work on second click in Info panel
Not sure if this is related to #219 or something else, but for About and Help in the Info panel, which draw on Wordpress pages with anchor links, it seems to require two clicks to get to where the anchor is pointing. The first click briefly seems to go there before returning to the top of the page.

abettermap:
- Fixes #223
- Fixes #219
- Fixes #119 (except iOS tile, can't get it to not be black, will try later but it's better than having React logo like we've had since the get-go!)
- Fixes #86


abettermap:
Couldn't get the app tile to work properly, it's supposed to have a white bg so the "L" isn't lost on dark screens, will have to look into it later but just wanted to get the logo into this branch since it's related-ish to welcome screen. also got the main page logo to switch to white when dark basemap set. neato!

rperlin-ela (approved):

PR #270, opened 2021-04-13
Post- post review

### LINKED ISSUE #218: Zooming back out from a point
Sometimes when zooming in on a point, it pulls you right back out to the default view. @abettermap notes that it only happens if you click on a point or go to it directly like from the URL, and the full data query for the table has not finished loading yet

### LINKED ISSUE #257: In-panel feedback form on mobile
Either make it look good in panel or scrap the in-panelness in favor of modal.

(broken out from #204)

### LINKED ISSUE #263: Alphabetizing languages where possible
It would be great to sort/alphabetize the language cards at the Explore/Neighborhood/Something level, the site cards at the Language Profile level, and the Additional Neighborhoods at the Language Profile (and likewise in the Site Details pop-out).

### LINKED ISSUE #269: Dark mode: improve map labels, logo transition
Show white text labels in map if World Region is not the symb.

### LINKED ISSUE #271: Scroll to panel top on route change
Sometimes the vertical scroll stays at the last place you were at, especially on mobile, when you change views/panels. Super annoying, should scroll to the top each time.￼

abettermap:
- Fixes #269
- Fixes #263
- Fixes #257
- Fixes #218
- Fixes #271


abettermap:
@rperlin-ela 

the "fix" for #218 is something that's part of some reallllly nasty convoluted messes which have to do with the table filters, clearing them, auto-zoom. so, not something i'm really inclined to include in prod comfortably as the very-last thing i do before docs. if you want to test it really hard (filtering, clearing filters, clicking map point as quickly as possible, reloading page from a point, etc.), let's go for it. but if you find anything fishy whatsoever regarding that stuff, let's scrap it. it's a small commit that's easy to undo and would only take a minute.

it's only wired up to prevent the unwanted zoom-to-full when a County, Neighb, or lang point is selected. doesn't matter if via click or page refresh. 

in other words if you start from /Census/puma/something/something via reload, you'll still get the accidental zoom. i figured those aren't advertised and aren't shareable though, so seemed like an edge case.

holler as soon as you have a conclusion. happy to scrap as i think it's part of a massive intertangled mess which could take a day or two to resolve.

abettermap:
not sure if i need to repeat this since we've got the PR flow down, but just make sure to review each of the issues closed above

abettermap:
forgot about icon too. let me know if you're not sure how to test the iOS tile icon. it's just "Add to Home Screen" from Safari share menu on iphone/pad.

abettermap:
sorry to be a last-minute-johnny here but i just fixed #271 as well. should work fine on mobile now.

abettermap:
deploy in a few minutes...

rperlin-ela (approved):
Didn't understand every bit in the PR but tested it all as best I could and can't find fault. And in under the gun with #271! Don't know if I fully grasped the issues with #218 but whatever I did the problem didn't recur, so I'd say let's go with it, tho your call in the end. Merge away!

abettermap:
Nice. 218 was when you click a point on the map as soon as possible after load, and then it kicks you back out to the full extent. Similarly if you started a page from a selected point, like hitting refresh or whatever, same thing.

Sounds like the Band-Aid worked though, merging.￼

PR #272, opened 2021-04-14
Update docs: improve README, init data structure

### LINKED ISSUE #8: Document Netlify, MB, GH
# Summary
Lots of moving parts, good to have docs for future reference. Some already in the Drive folder, some scattered in long GH issue/PR threads. so don't need redundancy.

## Services and tools to describe

- Netlify
- GitHub, GH Actions, Secrets, deploy/build/test processes
- Mapbox

## Info for each

- What it is
- Relationship to other tools/services
- Who manages it
- Summary of setup steps
- Links

## Where to put it?
Should this go in GitHub, like an `md` file or Wiki? Probably better in Google Doc since it shouldn't affect contributions or be of importance to non-members.

### LINKED ISSUE #107: Write bare minimum dev documentation
## Summary
The README theoretically covers the minimum steps for a dev to get up and running, but nothing about:

1. the data schema being a requirement, like the basis of the whole app working
1. configuring MB layers (not everyone will have Boundaries access like we do) or styles, etc.
1. WP content, which technically works as-is except it's hitting the languagemapping.org website. Not sure this is worth documenting for open-source though as a WP-driven things like an About page may not be of interest.

## For interested developers
If you want to use this project:

1. Create a language points layer in MB following the [`LangRecordSchema`](https://github.com/Language-Mapping/language-map/blob/20c3ecd49f9056bab79a253b339fb47cbdba1995/src/context/types.ts#L107-L125) schema in the _src/context/types.ts_ file.
1. Adjust the config in the config files as needed (not having MB Boundaries is probably going to wreck some things for you because it's tied to the lookup tables):
    image
1. Follow README instructions
1. If/when something breaks in this setup, please post an issue so we know what parts of the documentation to work on!

For item 2 above, you might be best off removing anything to do with `Boundaries` in _src/components/map/Map.tsx_.

abettermap:

- Fixes #107
- Fixes #8

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,10 @@
 
 # Languages of New York City Map
 
-[![Netlify Status](https://api.netlify.com/api/v1/badges/1f8bbbc8-4747-415e-bc63-b392256cedd5/deploy-status)](https://app.netlify.com/sites/languagemapping/deploys)
+[![Netlify
+Status](https://api.netlify.com/api/v1/badges/1f8bbbc8-4747-415e-bc63-b392256cedd5/deploy-status)](https://app.netlify.com/sites/languagemapping/deploys)
+
+## Background
 
 > This project will become an interactive map of New York City’s linguistic
 > diversity, using data collected by the Endangered Language Alliance (ELA).
@@ -16,7 +19,13 @@
 > toolkit documenting the workflows and tools used to enable others to undertake
 > language mapping in their own communities.
 
-## Setup and requirements
+## Additional resources
+
+View our [Google Drive documentation
+folder](https://drive.google.com/drive/u/1/folders/1OKVEZM_ossYa6E2djXpA4yTHCucu5D83)
+for additional resources, especially in regards to GIS workflows.
+
+## Dev setup and requirements
 
 This project was bootstrapped with [Create React
 App](https://github.com/facebook/create-react-app) with `--typescript`. In order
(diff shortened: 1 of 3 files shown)


abettermap:
@rperlin-ela

no deploy for this, but you can still view it [here](https://github.com/Language-Mapping/language-map/tree/c74dd3c374f573e153c9fc4233edc01eb5383a87) before we merge. some notes:

- readme won't make a ton of sense to you but give it a skim for typos i guess, and make sure it sorta makes sense and nothing is glaringly inaccurate.
- i gutted all the legacy junk, and made a _sample.env_ file for API keys if any devs use it. pretty common convention, and makes docs easier.
- i improved MB docs, placed in the _docs/data-structure.md_ file.

as for AT,  i didn't know how far to go w/this so that's where i stopped (in the data-structure file also). i found a way to grab the _entire_ data structure from AT in JSON format, and started whittling it down, but stopped for several reasons:

1. you can already access the info... in AT.
2. it's not terribly current in some cases, so that almost makes it worse than nothing at all in case the info is wrong/misleading.
3. i don't think the descrips i write for the cols and tables get read in the first place, so it seems futile to markdown-ify them.
4. the most important part is that the _code has what it needs_, and the major way i keep that intact is through the "types" file i mention. there's no connection w/AT, but it's as close as i can get to enforcing it.
5. if there are any changes to the data structure, that means i now have up to 4 places to document it: in the types, the in-code comments (potentially), AT, and this MD file. obviously not maintainable, and it's already pretty loose to begin with!
6. because there is no true enforced relationship between code and data, so the docs will be a mixed bag even if the code is good.
7. i think it would require a couple of days to document it all properly, so i'm holding out for CMS or similar in another scope. it would be more self-documenting i hope, but docs are a challenge to maintain either way!
8. some stuff that's in AT is not relevant whatsoever to the code. only the "types" are, but even some of those i'm kinda cowboying on the fly, and they're not in AT at all (hard to explain)

i left some efforts intact from my AT JSON dump (ha), and i do have my semi-whittled JSON attempts in case we want to expand the tables list (or gut it entirely, i don't care), but i'm just not sure how useful it would be since it's just redundant w/AT.

## The big things

...are the routing stuff, "types", and MB layers/fields/config, all of which i at least allude to in the docs. i also linked in readme to the Gdocs folder, in case you want to add more info there.

i know this was a pretty vague requirement in the first place so we can probably call it, but let me know if there's anything not living up to your immediate docs needs!

rperlin-ela (approved):
Thanks, much appreciated

PR #285, opened 2021-04-17
Resolve (hopefully) web-mercator assertion failure

### LINKED ISSUE #139: Error: @math.gl/web-mercator: assertion failed.
## Summary
I've seen this error _a lot_ and it seems to happen when the window is resized while the page is loading. Specific instances include rotating my phone or resizing the browser window.

## Solution
Not sure what to do with this yet. It's pretty detrimental though and is not being handled in the code, so probably start with a `try/catch` at a minimum. However, would want to make sure that the impacts of a broken "fly home" on load would not lead to MB not being able to "see" all the features and therefore pour a half-full glass of state in the global context.

It's probably okay though as the zoom-way-out, which happens before the fly-home, is where the state gets set. 

## Sentry info 

Sentry Issue: [LANGUAGE-MAPPING-X](https://sentry.io/organizations/endangered-language-alliance/issues/2012047220/?referrer=github_integration)

```
Error: @math.gl/web-mercator: assertion failed.
  at assert (../../../src/assert.js:6:15)
  at fitBounds (../../../src/fit-bounds.js:80:3)
  at fitBounds (../../../src/web-mercator-viewport.js:216:41)
  at utils (components/map/utils.ts:95:6)
  at flyHome (components/map/Map.tsx:370:5)
...
(5 additional frame(s) were not displayed)
```

abettermap:
- Fixes #139

### DIFF
--- a/src/components/map/utils.ts
+++ b/src/components/map/utils.ts
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/react'
 import { Map as MbMap, FillPaint, Layer, setRTLTextPlugin } from 'mapbox-gl'
 import { WebMercatorViewport } from 'react-map-gl'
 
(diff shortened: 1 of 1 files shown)


abettermap:
@rperlin-ela hopefully this fixes it. basically whenever there's a "zoom to bounds", be it a neighborhood, county, filtered lang points, or Zoom Home, i have a bit of "padding" set around the extent so that it's not crammed up against the edges of the map and potentially beneath the map controls, logo, etc. 

The problem occurs when (somehow) the screen is not large enough to accommodate both the padding and the bounds (e.g. of the polygon), if that makes sense. I still have no idea how users are triggering it, and I was only able to reproduce it by setting the map container to reeeeeally small, and THEN the problem showed up. I'm guessing it's something to do with resolution, like maybe these users have their screen set to enormously large or small, but it's soooo many people that I'm not sure whether such an edge-casey reason applies.

So now the "fix" is that I'm just checking whether the map height and width are larger than 2x the padding (left + right, or top + bottom). If not, I just fall back to NO padding. Not ideal but at least it won't break. I tested the deploy on as small a screen as Chrome Dev Tools would let me make, and no errors. Did the same in prod, w/same size, and our old friend showed up.

Merging!

abettermap:
as an added preventative measure, i wrapped that whole chunk of code in a `try/catch` block, which basically "tries" to do the thing, then aborts said thing if something goes wrong.

PR #309, opened 2026-04-28
Upgrade MUI to v5; replace material-table and react-swipeable-views

ankit0504:
## Summary

Phase 4 of the dependency-upgrade series, stacked on #306 (phase 3, React 17). This is the big one:

- **MUI v4 → v5** — `@material-ui/*` packages renamed to `@mui/*`, Emotion installed as the styling engine, JSS-style `makeStyles` kept via the official `@mui/styles` shim so the 91 v4-style files don't all need rewriting.
- **`material-table` removed.** It's unmaintained and never got a real MUI v5-compatible release. Replaced with `@tanstack/react-table` rendering MUI Table primitives.
- **`react-swipeable-views` removed.** Was used in exactly one place. Replaced with plain MUI Tabs.

## Why TanStack Table over the alternatives

The grid replacement was the biggest decision in this PR. Considered, in order:

| Option | Why ruled out |
| --- | --- |
| **MUI X Data Grid v6 community** | Tried first. Multi-column filtering is paywalled — `disableMultipleColumnsFiltering: true` is hard-coded in community v6. With material-table users routinely set multiple per-column filters, single-column-only is a real regression. |
| **MUI X Data Grid v5 community** | Free multi-column filtering, but last release was March 2023. v6/v7/v8 supersede it; v5 will not get fixes. Buys time, doesn't solve the problem. |
| **MUI X Data Grid v7 / v8** | Actively maintained but require React 18+. We're on React 17 (phase 3) and aren't doing the React 18 upgrade in this PR. Out of reach until a later phase. |
| **MUI X Pro/Premium** | ~$15/dev/mo. Buys multi-column filtering plus features we don't need (column pinning, tree data, etc.). Not worth it for a community project. |
| **`material-table@2.0.6`** | Actually MUI v5 + React 17 compatible (overlooked in initial review). Smallest diff. But abandoned since 2022, still depends on the abandoned `react-beautiful-dnd`, and locks us out of React 18+. We'd just have to redo this migration in phase 5. |
| **AG Grid Community** | Feature-rich and free, but ships its own design system (ag-theme-quartz) that won't match the rest of the MUI app without significant CSS work. Bundle is ~400kb. |
| **TanStack Table v8** ✅ | Headless (we render with MUI components ourselves), actively maintained, supports React 16/17/18+, no paywall now or later, no abandoned deps in the chain. Cost is a chunkier rewrite vs. Data Grid v5 — but it's a one-time cost that doesn't get re-spent at the next React/MUI bump. |

## Behavior changes vs. the old material-table

Preserved
- Per-column filter row under the headers (text inputs, select dropdowns for `Size`/`Status`, "Has media" checkbox for `Audio`/`Video`)
- Global search across visible columns
- Sortable columns including the custom Size order (Smallest → Largest, not alphabetic)
- Column show/hide menu
- CSV and PDF export of the currently-filtered, currently-sorted, currently-visible rows
- "View in map" and "Clear filters" buttons

Gone
- "Click cell to filter by its value" — was implemented via material-table's imperative `dataManager` API. Could be re-added on top of TanStack with a custom `onCellClick` later if missed.
- Sticky first/second columns. Material-table did this with absolute-positioned cells; we'd need similar custom CSS to bring it back.

## Test plan

- [x] Search box filters across all columns
- [x] Per-column filter inputs work (Language, Endonym, Country, etc.)
- [x] Size sort goes Smallest → Largest, not alphabetic
- [x] Audio/Video checkboxes filter to rows that have a URL
- [x] Status / Size dropdowns show only the lookup options
- [x] Column visibility menu shows/hides columns; CSV and PDF reflect the visible set
- [x] PDF export opens with non-Roman scripts rendered correctly
- [x] "View in map" pushes the filtered set into the map and navigates home
- [x] Pagination + page-size selector work
- [ ] Search tabs ("Search languages" / "Search locations") still toggle without the swipe gesture
- [x] Bottom nav, omnibox, dialog buttons, and breadcrumbs render fine with the v4→v5 styling shim

## What's next

This is one in a sequence of upgrade phases. Roughly:

- **Phase 5** — `react-router-dom` 5 → 6, `react-query` v2 → TanStack Query v4, `react-map-gl` 5 → 7, replace the deprecated `react-map-gl-geocoder`. Library swaps only — stays on React 17.
- **Phase 6** — React 17 → 19 (skipping 18 as a destination since most projects pass through it on the way to 19), and TanStack Query v4 → v5 in the same PR since v5 requires React 18+.

We may split phase 6 if React 19 surfaces a lot of concurrent-rendering / StrictMode behavior changes during testing.


abettermap:
@ankit0504 nice work, this is a big one! FYI I snuck in a commit to make the results modal fullscreen since we can no longer have the first two columns pinned, it's nice to have the extra room. 

Heads up, I'm not sure if this is due to another recent upgrade, but:

> Search tabs ("Search languages" / "Search locations") still toggle without the swipe gesture"

...is crashing in prod too. Will see if I can get a quick fix out. 

Also I see you are using Claude, so I'm generating a CLAUDE.md file, will include with this PR. Merging after a quick review.

abettermap:
I think we'll want to update the Claude file once these other PRs are merged.

vite.config.mts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+import svgr from 'vite-plugin-svgr'
+import tsconfigPaths from 'vite-tsconfig-paths'
+
+export default defineConfig({
+  plugins: [react(), svgr(), tsconfigPaths()],
+  envPrefix: ['REACT_APP_', 'VITE_'],
```
abettermap, line 8: we can get rid of this later, right?

abettermap (dismissed):

